PR #33, opened 2023-09-08
[SDP-852] Receiver registration: phone number input with countries dropdown

quietbits:
### What

Updated phone number input to show countries dropdown and validate the number for better UX. Jira ticket: https://stellarorg.atlassian.net/browse/SDP-852

image

image

### Why

It was not a good user experience to ask users to input the phone number in a specific format without any spaces, dashes, etc.

### Known limitations

N/A

#### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR's title starts with the name of the package, area, or subject affected by the change.

#### Thoroughness

#### Configs and Secrets

#### Release

#### Deployment


marcelosalloum (approved):
LGTM!

marcelosalloum:
I've tested it and it worked great. I haven't reviewed the code thoroughly though. Maybe you'd want a pass of @ceciliaromao before merging it.

marcelosalloum:
@quietbits: one small suggestion could be to set the starting country based on some IP information from the user, but that might be too much work, so I'm not sure it's worth it right now.

quietbits:
> @quietbits: one small suggestion could be to set the starting country based on some IP information from the user, but that might be too much work, so I'm not sure it's worth it right now.

@marcelosalloum This library might actually support that. I'll try that.

internal/htmltemplate/tmpl/receiver_register.tmpl
```diff
@@ -15,6 +15,9 @@
 
     <!-- Styles -->
     <link rel="stylesheet" href="/static/css/receiver_registration.css" />
+
+    <!-- Phone number input styles -->
+    <link rel="stylesheet" href="/static/css/intl-tel-input-v18.2.1.css" />
```
piyalbasu, line 20: I think it might be better to link to a cdn hosted version of the CSS and JS. You're probably trying to avoid additional roundtrips? It makes me a little nervous to manually add these lib files into the repo 😅 

Not sure how much of a lift it would be to use npm for this (I'm assuming significant), but I think if you're going to use an external lib, prob best to use either a cdn or a package manager

marcelosalloum: The security implication here is that we use Content-Security-Policy to allowlist specific domains where we will fetch resources. So if we want to use CDN, we just need to whitelist the address.

Is the CDN approach a recommendation for improved performance? Or is there another reason? If performance is the only factor, we could merge this PR and open a new one where we adopt the CDN, right?

quietbits: I don't think we'd gain anything in performance by using CDN. This will be the only page where these files are used, so caching locally vs from CDN wouldn't make much difference. It would be more convenient to upgrade to a newer version, though. By hosting the files ourselves, we're guaranteed that it will work if the page works. CDN might have a blackout, but this is probably unlikely.

I'd say let's merge this in and see how it goes. If we need to move to CDN, we can make the switch later. How do you feel about that, @piyalbasu @marcelosalloum ?

piyalbasu: I'm okay with merging this as is. My real recommendation would be to use npm or a package manager to handle external libs, but a CDN is probably less of a lift here. 

Locally hosting a versioned library just tends to be untenable in the long run and it is a bit of an antipattern. It adds friction when upgrading and adds unnecessary bloat to a codebase. Also, in my own experience with hosting code locally like this, someone always ends up directly editing the library files which can really cause headaches 😄 .

But none of these are dealbreakers to me. If you all are okay with taking on that cost, that's totally your call!

piyalbasu (approved):

ceciliaromao (commented):
LGTM! 🚀

ceciliaromao (approved):

internal/htmltemplate/tmpl/receiver_register.tmpl
marcelosalloum: I agree this is not ideal for application development, but considering this was meant to be a lightweight application that is used to get PII of recipients in a secure manner, we are weighing towards this kind of solution. That perception can change if we decide to increase the complexity of these pages, and if that's the case, we should definitely flag it.

> someone always ends up directly editing the library files

I agree this would be bad.

---
> a CDN is probably less of a lift in this situation, which is why I mentioned it.

Using CDN/cache would be a deployment-specific approach and every entity hosting the SDP would need to do that independently. We can recommend that in the README.

> My real recommendation would be to use npm or a package manager to handle external libs

Package management is probably a long-term solution and it's something that'd make sense if we decide to increase the complexity of this interface. I believe it could eventually become necessary, but it's not in our foreseeable future timeline yet. If @quietbits or @piyalbasu feel strongly about migrating to a react-app or another approach that uses npm package management, we can bring that up for discussion.

quietbits: We should make it a proper web app if it gets more complex. :100: I don't know what else we're planning to add, but this is probably the max we should do with the current setup.

PR #119, opened 2023-12-11
[SDP-922] events/serve: kafka setup

CaioTeixeira95:
### What

This PR adds the initial setup for the message broker's communication. This implementation uses `Apache Kafka` as the broker. The client Kafka used is the `segmention/kafka-go` package because it uses pure go and doesn't have any infra dependency to use it. The docker image used to run the Kafka locally is the `bitnami/kafka`.

### Why

We are going to use Kafka to run services asynchronously per tenant.

### Known limitations

N/A

#### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR's title starts with the name of the package, area, or subject affected by the change.

#### Thoroughness

#### Configs and Secrets

* [ ] No new CONFIG variables are required -OR- the new required ones were added to the helmchart's [`values.yaml`] file.
* [ ] No new CONFIG variables are required -OR- the new required ones were added to the deployments ([`pr-preview`], [`dev`], [`demo`], `prd`).
* [ ] No new SECRETS variables are required -OR- the new required ones were mentioned in the helmchart's [`values.yaml`] file.
* [ ] No new SECRETS variables are required -OR- the new required ones were added to the deployments ([`pr-preview secrets`], [`dev secrets`], [`demo secrets`], `prd secrets`).

#### Release

#### Deployment


internal/events/events.go
```diff
@@ -0,0 +1,80 @@
...
+			if errors.Is(err, io.EOF) {
+				log.Ctx(ctx).Warn("message broker returned EOF")
+				break
+			}
+			log.Errorf("error consuming: %s", err.Error())
+			return fmt.Errorf("consuming messages: %w", err)
+		}
+	}
+	return nil
+}
```
CaioTeixeira95, line 80: @marcelosalloum @marwen-abid What are your thoughts here? I let this only to discuss with you folks because if the consumer returns an error we will stop consuming messages, so I think we should think of something better here.

internal/events/kafka.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	log.Ctx(ctx).Infof("new message being processed: %s", msg.String())
+	for _, handler := range k.handlers {
+		if handler.CanHandleMessage(ctx, &msg) {
+			if err = handler.Handle(ctx, &msg); err != nil {
+				return fmt.Errorf("handler %s errored when handling message: %w", handler.Name(), err)
+			}
+		}
+	}
```
marcelosalloum, line 100: If some handlers succeed but one of the handlers fail we won't have an acknowledgment. This could be problematic for payments since it could cause payments to be re-processed.

Could you plz add a warning log for when the error is returned for a handler after one or more handlers have succeeded? We should make that edge case visible. Any thoughts here, @marwen-abid?

internal/serve/serve.go
```diff
@@ -90,6 +91,10 @@ type ServeOptions struct {
 	EnableMultiTenantDB             bool
 	tenantManager                   tenant.ManagerInterface
 	tenantRouter                    db.DataSourceRouter
+	EventProducer                   events.Producer
+	Brokers                         []string
+	Topics                          []string
+	ConsumerGroupID                 string
```
marcelosalloum, line 97: Does the server need anything Kafka-related? If not, we shouldn't be adding these fields to its object.

If it does need Kafka, we could be injecting a Kafka object, rather than injecting a few fields that would need to be tied together later. WDYT?

internal/events/events.go
```diff
@@ -0,0 +1,80 @@
...
+
+func Consume(ctx context.Context, consumer Consumer) error {
+	log.Ctx(ctx).Info("starting consuming messages...")
+	for {
+		if err := consumer.ReadMessage(ctx); err != nil {
+			if errors.Is(err, io.EOF) {
+				log.Ctx(ctx).Warn("message broker returned EOF")
+				break
+			}
+			log.Errorf("error consuming: %s", err.Error())
```
marcelosalloum, line 75: Can you use the context when logging, plz?

internal/events/events.go
```diff
@@ -0,0 +1,80 @@
...
+type Consumer interface {
+	RegisterEventHandler(ctx context.Context, eventHandlers ...EventHandler) error
+	ReadMessage(ctx context.Context) error
+	Close() error
+}
+
+func Consume(ctx context.Context, consumer Consumer) error {
+	log.Ctx(ctx).Info("starting consuming messages...")
+	for {
+		if err := consumer.ReadMessage(ctx); err != nil {
```
marcelosalloum, line 70: Usually, you'd want to add ways to break this for loop. For instance, when the context is cancelled or when the OS sends a SIGINT.

We're currently doing that in the TSS, so you can replicate that:
https://github.com/stellar/stellar-disbursement-platform-backend/blob/36ed4b5f9b88968ca12bf5dbd1ab16883541aff0/internal/transactionsubmission/manager.go#L216-L222

internal/events/events.go
marcelosalloum: I'd say we need to call the crash tracker in this case. But I defer the final answer to @marwen-abid, since he's more experienced in handling Kafka errors.

marcelosalloum (commented):
Looking great!

I've left a few comments below.

cmd/serve.go
```diff
@@ -309,6 +310,29 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
...
+			CustomSetValue: cmdUtils.SetConfigOptionStringList,
+			Required:       true,
+		},
+		{
+			Name:      "consumer-group-id",
+			Usage:     "Message Broker Consumer Group ID.",
+			OptType:   types.String,
+			ConfigKey: &serveOpts.ConsumerGroupID,
+			Required:  true,
+		},
```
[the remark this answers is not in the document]
CaioTeixeira95: This is for configuring the producer/consumer. What do you think?

internal/events/kafka.go
```diff
@@ -0,0 +1,115 @@
...
+	reader   *kafka.Reader
+}
+
+func NewKafkaEventManager(brokers []string, consumerTopics []string, consumerGroupID string) (*KafkaEventManager, error) {
+	k := KafkaEventManager{}
+
+	writer := kafka.NewWriter(kafka.WriterConfig{
+		Brokers:      brokers,
+		Balancer:     &kafka.RoundRobin{},
+		RequiredAcks: -1,
```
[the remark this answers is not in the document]
CaioTeixeira95: Good call. Here is the producer's config. It means it will ensure the message was written on all brokers/partitions before returning. If it wasn't written, it returns an error. For the consumer, we use the consume message and after processing it we ACK it.

The library we use ([kafka-go](https://github.com/segmentio/kafka-go)), we have two methods to consume messages:

- `ReadMessage` - Reads the message and commits the offset.
- `FetchMessage` - Reads the message and does not commit the offset. We must use the `CommitMessages` method to commit it explicitly.

The approach I've chosen was the `FetchMessage` one.

internal/events/kafka.go
CaioTeixeira95: Got it. So the approach is to only log the error and not return it. Is that correct?

internal/serve/serve.go
CaioTeixeira95: I don't think we need it! I'm going to remove it.

cmd/serve.go
```diff
@@ -424,6 +448,14 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
...
 			serveOpts.AnchorPlatformAPIService = apAPIService
 
+			// Kafka (background)
+			kafkaEventManager, _ := events.NewKafkaEventManager(serveOpts.Brokers, serveOpts.Topics, serveOpts.ConsumerGroupID)
+			kafkaEventManager.RegisterEventHandler(ctx, &events.PingPongEventHandler{})
+			defer kafkaEventManager.Close()
+
+			go events.Consume(ctx, kafkaEventManager)
+			serveOpts.EventProducer = kafkaEventManager
+
```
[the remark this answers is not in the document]
CaioTeixeira95: I'm sorry, I've improved this! Could you please take a look again?

internal/events/kafka.go
```diff
@@ -0,0 +1,115 @@
...
+type KafkaEventManager struct {
+	handlers []EventHandler
+	writer   *kafka.Writer
+	reader   *kafka.Reader
+}
+
+func NewKafkaEventManager(brokers []string, consumerTopics []string, consumerGroupID string) (*KafkaEventManager, error) {
+	k := KafkaEventManager{}
+
+	writer := kafka.NewWriter(kafka.WriterConfig{
```
marwen-abid, line 22: `kafka.newWriter` and `kafka.WriterConfig` seem to be deprecated. I think the following should work instead: 
```
writer := &kafka.Writer{
  Addr:     kafka.TCP(brokers...),
  Balancer: &kafka.RoundRobin{},
  RequiredAcks: -1,
}
```

cmd/serve.go
```diff
@@ -424,6 +448,21 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 			}
 			serveOpts.AnchorPlatformAPIService = apAPIService
 
+			// Kafka (background)
+			kafkaEventManager, err := di.NewKafkaEventManager(ctx, serveOpts.Brokers, serveOpts.Topics, serveOpts.ConsumerGroupID, &events.PingPongEventHandler{})
```
marwen-abid, line 452: I believe we should separate `KafkaConsumer` and `KafkaProducer`. `KafkaEventManager` seems to mix both concepts even though they're pretty separate. 
## KafkaProducer: 
- requires just brokers and topic. 
- is going to be used all over the code, whenever we need to produce a message. 
- lightweight. 
- we only need one instance. 

## KafkaConsumer
- we need multiple instances 
- it only needs to be defined on startup. 
- heavyweight (requires looping). 

This is just example to illustrate what I'm thinking 
```go
// Kafka Producer.
serveOpts.KafkaProducer, err := di.NewKafkaProducer(ctx, serveOpts.Brokers)
// error handling
defer serveOpts.KafkaProducer.Close()

// Kafka Consumers
// 1. Consumer for `sdp.payments.payments_submitted` topic
paymentsConsumer, err :=  di.NewKafkaConsumer(ctx, serveOpts.Brokers, serveOpts.TopicPaymentSubmitted, serveOpts.ConsumerGroupID, &handlers.SubmittedPaymentEventHandler{})
// error handling
defer paymentSubmittedConsumer.Close()

go func() {
   err := events.Consume(ctx, paymentSubmittedConsumer)
}

// 2. Consumer for `sdp.payments.payment_status_updated` topic
paymentStatusUpdatedConsumer, err :=  di.NewKafkaConsumer(ctx, serveOpts.Brokers, serveOpts.TopicPaymentStatusUpdated, serveOpts.ConsumerGroupID,
&handlers.PaymentStatusUpdatedEventHandler{})
// error handling
defer paymentStatusUpdatedConsumer.Close()

go func() {
    err := events.Consume(ctx, paymentStatusUpdatedConsumer)
}
```

We can talk about it before you make the change if you think this won't work.

internal/events/kafka.go
```diff
@@ -0,0 +1,115 @@
...
+		Brokers:      brokers,
+		Balancer:     &kafka.RoundRobin{},
+		RequiredAcks: -1,
+	})
+
+	reader := kafka.NewReader(kafka.ReaderConfig{
+		Brokers:     brokers,
+		GroupID:     consumerGroupID,
+		GroupTopics: consumerTopics,
```
marwen-abid, line 31: I'm not sure about using `GroupTopics`. I never used it before so we have to look into this. 

How would consumption work ? Would we round robin between these topics and get one message from each ? In that case we may be delaying important messages in favor of less important messages. 

I think we may need to switch to an implementation that handles one topic per reader/consumer. We can have multiple handlers per consumer (e.g. `sdp.payments.payment_status_updated` would be handled both by `PatchAnchorPlatformEventHandler` and `UpdatePaymentStatusEventHandler` ).

marwen-abid (commented):
This looks great! Really good work wiring this whole thing together.

internal/events/kafka.go
CaioTeixeira95: From the library docs:

> GroupTopics allows specifying multiple topics, but can only be used in combination with GroupID, as it is a consumer-group feature. As such, if GroupID is set, then either Topic or GroupTopics must be defined.

But, I believe we can fall into the problem you said, delaying important messages in favor of less important ones. I can change the implementation to separate the consumer from the producer, just like you [suggested here](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119#discussion_r1424913197). I think this approach will make things easy to use and test.

Thank you so much for your comment!

CaioTeixeira95: Changed: [87c250d](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/87c250d9d3feb26893f037c7735c336056f96743)

cmd/serve.go
CaioTeixeira95: changed: [87c250d](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/87c250d9d3feb26893f037c7735c336056f96743)

internal/events/kafka.go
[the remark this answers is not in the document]
CaioTeixeira95: changed: [f32b758](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/f32b7586fea76ccaf53b7ec2ac42e3c79c2357fd)

internal/events/kafka.go
CaioTeixeira95: changed: [90c18a3](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/90c18a35c2383bd3bc700f326c21b0e3d33c6fd6)

internal/events/events.go
```diff
@@ -0,0 +1,79 @@
...
+type Producer interface {
+	WriteMessages(ctx context.Context, messages ...Message) error
+	Close() error
+}
+
+type Consumer interface {
+	RegisterEventHandler(ctx context.Context, eventHandlers ...EventHandler) error
+	ReadMessage(ctx context.Context) error
+	Close() error
+}
```
marwen-abid, line 66: nit: A couple of things with this interface: 
1. `ReadMessage()` should probably be `ReadAndHandleMessage()`. Because we're also invoking the handler. 
2. `ReadMessage()` doesn't need to be exposed, it should be an implementation detail. 
3. Consume(..) should probably be on this interface. 

So my final interface would look like : 
```go
type Consumer interface {
	RegisterEventHandler(ctx context.Context, eventHandlers ...EventHandler) error
	Consume(ctx context.Context, consumer Consumer, crashTracker crashtracker.CrashTrackerClient)
	Close() error
}
```
This is how it would be used : 
```go
pingPongConsumer := events.NewKafkaConsumer(brokers, "ping-pong", consumerGroupID)
err := pingPongConsumer.RegisterEventHandler(ctx, events.PingPongEventHandler{})
defer pingPongConsumer.Close()
go pingPongConsumer.Consume(ctx, crashTrackerClient)
```

Based on this contract `Read()` or `ReadAndHandle()` doesn't need to be exposed.

internal/events/events.go
marwen-abid: @CaioTeixeira95 This is a good suggestion I think. We should add the cancellation.

internal/dependencyinjection/event_manager.go
```diff
@@ -0,0 +1,30 @@
+package dependencyinjection
```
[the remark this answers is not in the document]
CaioTeixeira95: Yeah, we need this because of the `TestServe` since it will be the same object, the test will pass. Otherwise, they will have different memory addresses, and the test fails.

internal/events/events.go
CaioTeixeira95: The `Consume` method should not be something from the `Consumer` interface, IMO. The main problem is that we'll reimplement the same logic for every consumer on this method. But, I agree that we should add the `crashTracker` on the Consumer. But it can be something as an attribute, not a function argument.

internal/events/events.go
CaioTeixeira95: changed: [e3e7307](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/e3e7307805afd96364e340a49ba58919bec502c4)

internal/events/events.go
CaioTeixeira95: changed: [9ee7d22](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/9ee7d22a1d0d32495233dfe75be43cc790dba533)

internal/dependencyinjection/event_manager.go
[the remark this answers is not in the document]
CaioTeixeira95: changed: [9ee7d22](https://github.com/stellar/stellar-disbursement-platform-backend/pull/119/commits/9ee7d22a1d0d32495233dfe75be43cc790dba533)

marwen-abid (approved):
This looks great! Thank you so much @CaioTeixeira95  🚀 🚀 🚀

PR #122, opened 2023-12-11
[SDP-996] Make `POST /assets` idempotent

marcelosalloum:
### What

Make `POST /assets` idempotent.

### Why

The previous implementation was causing a hassle with a partner that had seeded their database before using the frontend application.

#### PR Structure

#### Thoroughness

#### Configs and Secrets

#### Release

#### Deployment

* [x] Does the deployment work after merging?

### DIFF
--- a/internal/data/assets.go
+++ b/internal/data/assets.go
@@ -123,7 +123,7 @@ func (a *AssetModel) GetAll(ctx context.Context) ([]Asset, error) {
 	return assets, nil
 }
 
-// Insert is idempotent and returns a new asset if it doesn't exist or the existing one if it does, clrearing the
+// Insert is idempotent and returns a new asset if it doesn't exist or the existing one if it does, clearing the
 // deleted_at field if it was marked as deleted.
 func (a *AssetModel) Insert(ctx context.Context, sqlExec db.SQLExecuter, code string, issuer string) (*Asset, error) {
 	const query = `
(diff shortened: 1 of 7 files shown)


internal/data/assets.go
```diff
@@ -123,18 +123,25 @@ func (a *AssetModel) GetAll(ctx context.Context) ([]Asset, error) {
...
+				($1, $2)
+			ON CONFLICT (code, issuer) DO
+			UPDATE SET
+				deleted_at = NULL
+			WHERE
+				assets.deleted_at IS NOT NULL
+			RETURNING *
+		)
+		SELECT * FROM upsert_asset
+		UNION ALL  -- // The UNION statement is applied to prevent the updated_at field from beong autoupdated when the asset already exists.
```
marwen-abid, line 143:
```suggestion
		UNION ALL  -- // The UNION statement is applied to prevent the updated_at field from being autoupdated when the asset already exists.
```

internal/data/assets.go
```diff
@@ -123,18 +123,25 @@ func (a *AssetModel) GetAll(ctx context.Context) ([]Asset, error) {
...
+		SELECT * FROM assets WHERE code = $1 AND issuer = $2 AND NOT EXISTS (SELECT 1 FROM upsert_asset);
```
marwen-abid, line 144: I don't see why the following query wouldn't work here ? What's the edge case we're covering with the extra WITH & UNION. 
```
INSERT INTO assets (code, issuer)
VALUES ($1, $2)
ON CONFLICT (code, issuer) DO
UPDATE SET
    deleted_at = NULL
WHERE
    assets.deleted_at IS NOT NULL
RETURNING *;
```

marwen-abid (approved):
LGTM. Just a couple of typos and a question

internal/data/assets.go
marcelosalloum: Let's analyze how the INSERT statement works for three different use cases:
1. **The asset does not exist in the DB**. In that case, it will be inserted without a conflict and all fields will be returned.
2. **The asset exists in the database but it's marked as deleted**. In that case, the conflict will be raised, the deleted_at field will be cleared and all the fields will be returned.
3. **The asset exists in the database but it's not marked as deleted**. In that case, nothing is returned from this statement because of the WHERE clause. It's sad, I know.

The two options I could find to make the 3rd use case return the object was either:
a. setting the deleted_at to NULL every time, which would call the updated_at trigger, and I didn't want that.
b. using this union statement to compensate for that.

There's a post about this problem at https://stackoverflow.com/questions/35949877/how-to-include-excluded-rows-in-returning-from-insert-on-conflict/35953488#35953488.

PR #128, opened 2023-12-15
SDP-980 Remove CORS from tenant configuration

marwen-abid:
Matching kube PR https://github.com/stellar/kube/pull/1859

### TL;DR 
Experimented with multiple solutions to enable CORS in multi-tenancy. We don't need to have a CORS configuration per tenant as we originally thought, we can use a generic wildcard config thanks to this library https://github.com/rs/cors. 

### Explanation
We originally thought that we will need to define one cors configuration per tenant as CORS headers either allow wildcard `*` or a strict domain e.g. `https://redcorp.mtn-sdp-backend-dev.stellar.org` , `https://bluecorp.mtn-sdp-backend-dev.stellar.org` 

The initial implementation I tested relied on that concept 
```go
func CorsMiddleware(corsAllowedOrigins []string) func(http.Handler) http.Handler {
	return func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {

			allowedOrigins := corsAllowedOrigins

			// Extract tenant from context if possible
			ctx := r.Context()
			t, err := tenant.GetTenantFromContext(ctx)
			if err != nil {
				if !errors.Is(err, tenant.ErrTenantNotFoundInContext) {
					log.Ctx(ctx).Errorf("Error getting tenant from context: %s", err)
				}
			} else {
				allowedOrigins = append(allowedOrigins, t.CORSAllowedOrigins...)
			}

			// Dynamically set the CORS allowed origins based on the tenant
			cors := cors.New(cors.Options{
				AllowedOrigins: allowedOrigins,
				AllowedHeaders: []string{"*"},
				AllowedMethods: []string{"GET", "PUT", "POST", "PATCH", "DELETE", "HEAD", "OPTIONS"},
			})

			cors.Handler(next).ServeHTTP(w, r)
		})
	}
}
```

This works fine for most cases but requires the CORS middleware to be tenant-aware. Which means that middleware needs to be injected differently & conditionally depending on a tenant's existence. 

**CORS library**
https://github.com/rs/cors
When experimenting with other solutions before committing to the one described above, I found out that the CORS library we use allows for wildcard domains. Even though the header doesn't accept a wildcard, the library does a treatment based on the configuration and the origin header and sets the proper strict cors header. 
From the documentation : 
```
// AllowedOrigins is a list of origins a cross-domain request can be executed from.
// If the special "*" value is present in the list, all origins will be allowed.
// An origin may contain a wildcard (*) to replace 0 or more characters
// (i.e.: http://*.domain.com). Usage of wildcards implies a small performance penalty.
// Only one wildcard can be used per origin.
```
This solves our problems.

### DIFF
--- a/db/migrate_test.go
+++ b/db/migrate_test.go
@@ -125,11 +125,11 @@ func TestMigrate_downApplyOne_Tenant_migrations(t *testing.T) {
 
 	n, err := Migrate(db.DSN, migrate.Up, 2, tenantmigrations.FS, StellarMultiTenantMigrationsTableName)
 	require.NoError(t, err)
-	require.Equal(t, 1, n)
+	require.Equal(t, 2, n)
 
-	n, err = Migrate(db.DSN, migrate.Down, 1, tenantmigrations.FS, StellarMultiTenantMigrationsTableName)
+	n, err = Migrate(db.DSN, migrate.Down, 2, tenantmigrations.FS, StellarMultiTenantMigrationsTableName)
 	require.NoError(t, err)
-	require.Equal(t, 1, n)
+	require.Equal(t, 2, n)
 
 	ids := []string{}
 	err = dbConnectionPool.SelectContext(ctx, &ids, fmt.Sprintf("SELECT id FROM %s", StellarMultiTenantMigrationsTableName))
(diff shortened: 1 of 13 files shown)


CaioTeixeira95 (approved):
LGTM

PR #153, opened 2024-01-19
chore - expose admin port for the service

marwen-abid:
### What

We need to expose the admin port 8003 in order to be able to access the admin from outside of the container. 

This doesn't create an Ingress to the admin API.


marcelosalloum (approved):
LGTM

CaioTeixeira95 (approved):

PR #234, opened 2024-03-25
[SDP-1119] [Refactor] refactoring tenant provisioning

marwen-abid:


stellar-multitenant/pkg/internal/provisioning/manager.go
```diff
@@ -49,238 +59,261 @@ func rollbackTenantDataSetupErrors() []error {
...
-		if err != nil {
-			if errors.Is(err, signing.ErrUnsupportedCommand) {
-				log.Ctx(ctx).Warnf(
-					"Account provisioning not needed for distribution account signature client type %s: %v",
-					m.SubmitterEngine.DistAccountSigner.Type(), err)
-			} else {
-				return nil, fmt.Errorf("%w: provisioning distribution account: %w", ErrProvisionTenantDistributionAccountFailed, err)
-			}
-		}
+func (m *Manager) handleProvisioningError(ctx context.Context, name string, err error, t *tenant.Tenant) error {
```
marwen-abid, line 99: Do we need the name here as a parameter if we also have the tenant ?

stellar-multitenant/pkg/internal/provisioning/manager.go
```diff
@@ -49,238 +59,261 @@ func rollbackTenantDataSetupErrors() []error {
...
-					provisioningErr = fmt.Errorf("%w. [additional errors]: %s: %w", provisioningErr, deleteDistributionKeyErrPrefixMsg, deleteDistributionAccFromVaultErr)
-					log.Ctx(ctx).Errorf("%s: %v", deleteDistributionKeyErrPrefixMsg, deleteDistributionAccFromVaultErr)
-				}
-
-				log.Ctx(ctx).Errorf("distribution account cleanup successful")
-				break
-			}
-		}
+		log.Ctx(ctx).Errorf("tenant %s deleted", name)
+	}
```
marwen-abid, line 133: I think we can merge `rollbackTenantCreationErrors()` and `rollbackTenantSchemaErrors()` because they both cover the same set of errors. 

Because they both cover the same errors 
```go 
func rollbackTenantCreationErrors() []error {
	return []error{ErrUpdateTenantFailed, ErrProvisionTenantDistributionAccountFailed, ErrTenantDataSetupFailed, ErrTenantSchemaFailed}
}

func rollbackTenantSchemaErrors() []error {
	return []error{ErrUpdateTenantFailed, ErrProvisionTenantDistributionAccountFailed, ErrTenantDataSetupFailed, ErrTenantSchemaFailed}
}
```

stellar-multitenant/pkg/internal/provisioning/manager.go
ziyliu: nope, thanks for calling it out

PR #249, opened 2024-04-04
[SDP-1100] middleware/stellar-multitenant/serve/cmd: add default tenant

CaioTeixeira95:
### What

This PR adds the default tenant feature. A new environment variable was added `ENABLE_DEFAULT_TENANT` and a new column was added to the `public.tenants` table `is_default`.

It's now possible to set the env var `SINGLE_TENANT_MODE` to `true` and then configure a tenant to be the default tenant by setting the `is_default` to `true` as well.

After this, every request will be resolved to the default tenant regardless of the `Sdp-Tenant-Name` request header or the `Hostname prefix`. This is useful when configuring the SDP to run with only one tenant.

### Why

Make the setup for a single-tenant configuration easier.

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Configs and Secrets

#### Release

#### Deployment

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -326,7 +326,7 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 		},
 		{
 			Name:        "enable-default-tenant",
-			Usage:       "This option enables the Default Tenant feature. In case the multi-tenant management is not required, this options bypass the tenant resolution by always resolving to the default tenant configured in the database.",
+			Usage:       "This option enables the Default Tenant feature. In the case where multi-tenancy is not required, this options bypasses the tenant resolution by always resolving to the default tenant configured in the database.",
 			OptType:     types.Bool,
 			ConfigKey:   &serveOpts.EnableDefaultTenant,
 			FlagDefault: false,
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -325,10 +325,10 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 			Required:    true,
 		},
 		{
-			Name:        "enable-default-tenant",
-			Usage:       "This option enables the Default Tenant feature. In the case where multi-tenancy is not required, this options bypasses the tenant resolution by always resolving to the default tenant configured in the database.",
+			Name:        "single-tenant-mode",
+			Usage:       "This option enables the Single Tenant Mode feature. In the case where multi-tenancy is not required, this options bypasses the tenant resolution by always resolving to the default tenant configured in the database.",
 			OptType:     types.Bool,
-			ConfigKey:   &serveOpts.EnableDefaultTenant,
+			ConfigKey:   &serveOpts.SingleTenantMode,
 			FlagDefault: false,
 		},
 	}
@@ -610,7 +610,7 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 			go serverService.StartMetricsServe(metricsServeOpts, &serve.HTTPServer{})
 
 			log.Ctx(ctx).Info("Starting Tenant Server...")
-			adminServeOpts.EnableDefaultTenant = serveOpts.EnableDefaultTenant
+			adminServeOpts.SingleTenantMode = serveOpts.SingleTenantMode
 			go serverService.StartAdminServe(adminServeOpts, &serveadmin.HTTPServer{})
 
 			// Starting Application Server
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -324,6 +324,13 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 			FlagDefault: 3,
 			Required:    true,
 		},
+		{
+			Name:        "enable-default-tenant",
+			Usage:       "This option enables the Default Tenant feature. In case the multi-tenant management is not required, this options bypass the tenant resolution by always resolving to the default tenant configured in the database.",
+			OptType:     types.Bool,
+			ConfigKey:   &serveOpts.EnableDefaultTenant,
+			FlagDefault: false,
+		},
 	}
 
 	// crash tracker options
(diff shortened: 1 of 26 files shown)


stellar-multitenant/pkg/tenant/manager.go
```diff
@@ -213,6 +235,11 @@ func (m *Manager) UpdateTenantConfig(ctx context.Context, tu *TenantUpdate) (*Te
 		log.Ctx(ctx).Warnf("distribution account for tenant id %s updated to %s", tu.ID, *tu.DistributionAccount)
 	}
 
+	if tu.IsDefault != nil {
```
marwen-abid, line 238: I believe we should not update `IsDefault` in the same way as we update other fields.

When a `POST` or `PATCH` call attempts to update `is_default` we should consider doing the following: 
* **At the handler level**: 
   * Fail this call if `enable-default-tenant=false`. You need to enable default tenant feature to be able to update this flag. 
   * Only accept `true` as a valid value. We always want to have **one** valid default tenant. So if a user ever wants to turn off default for a tenant, it should turn it on for another tenant. 
* **In `UpdateTenantConfig`**: this call would set `is_default = true` for the current tenant, and would set `is_default = false` for all other tenants. 

wdyt ?

marwen-abid (approved):
LGTM!

stellar-multitenant/pkg/tenant/manager.go
CaioTeixeira95: That sounds good! I'll work on it.

cmd/serve.go
```diff
@@ -324,6 +324,13 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 			FlagDefault: 3,
 			Required:    true,
 		},
+		{
+			Name:        "enable-default-tenant",
+			Usage:       "This option enables the Default Tenant feature. In the case where multi-tenancy is not required, this options bypasses the tenant resolution by always resolving to the default tenant configured in the database.",
+			OptType:     types.Bool,
+			ConfigKey:   &serveOpts.EnableDefaultTenant,
+			FlagDefault: false,
+		},
```
marcelosalloum, line 333: 🤔  "enable default tenant" means that when none is provided, it will be resolved to a default one. But the implementation seems to be "disabling multi tenancy". When I reviewed [the ticket](https://stellarorg.atlassian.net/browse/SDP-1100), it was not clear to me if the idea was to ignore/disable any tenant that doesn't have `is_default=true`.

marcelosalloum (commented):
I need more context in order to review this. I shared a question below 👇

cmd/serve.go
CaioTeixeira95: Changed to `single-tenant-mode`

dev/docker-compose-sdp-anchor.yml
```diff
@@ -49,6 +49,7 @@ services:
       INSTANCE_NAME: "SDP Testnet on Docker"
       TENANT_XLM_BOOTSTRAP_AMOUNT: 5
       DISTRIBUTION_SIGNER_TYPE: ${DISTRIBUTION_SIGNER_TYPE:-DISTRIBUTION_ACCOUNT_ENV}
+      ENABLE_DEFAULT_TENANT: "false"
```
marcelosalloum, line 52: 👋  You forgot to rename the env here

internal/anchorplatform/sep24_auth_middleware_test.go
```diff
@@ -697,3 +700,37 @@ func Test_SEP24HeaderTokenAuthenticateMiddleware(t *testing.T) {
 		require.Contains(t, buf.String(), "missing home domain in the token claims")
 	})
 }
+
+func Test_getCurrentTenant(t *testing.T) {
```
marcelosalloum, line 704: I'm missing the successful paths in this test.

internal/serve/middleware/middleware.go
```diff
@@ -285,16 +285,31 @@ func CSPMiddleware() func(http.Handler) http.Handler {
...
+					switch {
+					case errors.Is(err, tenant.ErrTenantDoesNotExist):
+						httperror.Unauthorized("No default Tenant configured", err, nil).Render(rw)
+					case errors.Is(err, tenant.ErrTooManyDefaultTenants):
+						httperror.Unauthorized("Too many default tenants configured", err, nil).Render(rw)
+					default:
+						httperror.InternalError(ctx, "", fmt.Errorf("error getting default tenant: %w", err), nil).Render(rw)
+					}
+					return
+				}
```
marcelosalloum, line 307: 4xx are usually caused by a problem with the request/client, while 5xx refers to problems in the server and server data.

In this case, I don't see why we would return Unauthorized. a 500 for any of these cases would make more sense to me since there's nothing the client can do to get a 2xx response.

stellar-multitenant/pkg/internal/httphandler/tenants_handler.go
```diff
@@ -143,3 +146,45 @@ func (t TenantsHandler) Patch(w http.ResponseWriter, r *http.Request) {
 
 	httpjson.RenderStatus(w, http.StatusOK, tnt, httpjson.JSON)
 }
+
+func (t TenantsHandler) SetDefault(rw http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	if !t.SingleTenantMode {
+		log.Ctx(ctx).Warnf("An attempt of set a default tenant was made but ENABLE_DEFAULT_TENANT is set to `false`")
```
marcelosalloum, line 154: 👋  You forgot to rename the env here

marcelosalloum (approved):
LGTM! Plz, feel free to merge it after addressing the comments below.

internal/serve/middleware/middleware.go
CaioTeixeira95: I'll change them to 500 but I'll keep the messages separated to make the debug easier

marcelosalloum (approved):

PR #256, opened 2024-04-12
SDP-1116 Support External_ID DRAFT PR (See TODO). reece

reecexlm:
### What

This change will be deployed in a branch with a PR Preview pipeline  to support Vibrant Assist in testing integration of supporting external_id in SDP registration deep link and sep24 post interactive deposit for partner integration.

The receiver’s invitation DeepLink should include the field external_id, containing the ID of that receiver_wallet

This code will not be merged for now. We should decide if we want to adopt this external-id field or not.

TBH, it could remove the need for an OTP step on our end, but we should discuss if we want that.

We need to validate the decision with @Nando Vieira  before concluding if this will be part of the official implementation, or if an optional field.

added some additional unit tests and verified all existing tests are passing.

### Why

To support reconciliation of vibrant client (mobile phone) with SDP Host Customer ID by using an external_id field in the deeplink sent from the sdp host (SDP registration deep link). then this will be passed within SEP24 claims for reconciliation.

### Known limitations / TODO
TBD - not yet part of this PR: When validating the wallet-registration, verify if the SEP24 incoming info contains the field: customer_id=external_id:hash(phone_number) Use the external_id to find the user in our database Hash the user’s phone number and compare it with the incoming hashed phone number It’s ok to send an informative response, so Vibrant can understand this well.

#### PR Structure

* [x ] This PR has a reasonably narrow scope
* [x ] This PR title and description are clear enough for anyone to review it.
* [ ] This PR does not mix refactoring changes with feature changes (split into two PRs otherwise).

#### Thoroughness

* [ ] This PR adds tests for the new functionality or fixes.
* [ ] This PR contains the link to the Jira ticket it addresses.

#### Configs and Secrets
there is a new configuration option useExternalID (true/false) to make this optional.  This has not yet been updated to helm charts yet until we determine if we plan to include this support in our standard release.

#### Release
not for release at this time.  only for testing with Vibrant.

#### Deployment
Will be deployed only in a pr-preview pipeline at this time.


internal/data/assets_test.go
```diff
@@ -735,6 +742,7 @@ func Test_GetAssetsPerReceiverWallet(t *testing.T) {
 					ID:          receiverY.ID,
 					Email:       receiverY.Email,
 					PhoneNumber: receiverY.PhoneNumber,
+					ExternalID: receiverY.ExternalID,
```
marcelosalloum, line 745: I think your IDE is not indenting this correctly. Maybe this is the old spaces vs tab indentation issue.

internal/serve/serve.go
marcelosalloum, line 1: You will need to propagate the UseExternalID config to the handler responsible for `/wallet-registration/verification`, injecting it in:

https://github.com/stellar/stellar-disbursement-platform-backend/blob/977e0c359b76a61c6167fa30aff5b2e6912ed283/internal/serve/serve.go#L435-L441

internal/serve/httphandler/verifiy_receiver_registration_handler.go
```diff
@@ -108,6 +108,12 @@ func (v VerifyReceiverRegistrationHandler) processReceiverVerificationPII(
 	now := time.Now()
 	truncatedPhoneNumber := utils.TruncateString(receiver.PhoneNumber, 3)
 
+	// Use the external_id associated with the receiver if it exists
+	if receiver.ExternalID != "" {
+		// Logic to handle the verification associated with the external_id
+		// This may include comparing the hashed phone number with the one associated with external_id
+	}
+
```
marcelosalloum, line 116: You'll probably want to do the externalID validation (and the hashed_phone_number validation) on or before the `Step 3` below:

https://github.com/stellar/stellar-disbursement-platform-backend/blob/977e0c359b76a61c6167fa30aff5b2e6912ed283/internal/serve/httphandler/verifiy_receiver_registration_handler.go#L136-L153

cmd/serve.go
marcelosalloum, line 1: You'll need to inject this new config in `WithSendReceiverWalletsSMSInvitationJobOption`:
https://github.com/stellar/stellar-disbursement-platform-backend/blob/977e0c359b76a61c6167fa30aff5b2e6912ed283/cmd/serve.go#L100-L107

marcelosalloum (commented):
This is looking pretty good, congrats!

I've shared some thoughts below on:
- some connection points that are missing (to make things easier for you)
- my opinion on where to do the final verification when validating the user in the SEP24 request.

Also, a nitpicky thing: indentation is looking a bit weird in some places. It looks like the old problem of spaces vs tabs indentation.

internal/services/send_receiver_wallets_invite_service.go
```diff
@@ -283,13 +290,14 @@ func (s SendReceiverWalletInviteService) shouldSendInvitationSMS(ctx context.Con
...
 
-func NewSendReceiverWalletInviteService(models *data.Models, messengerClient message.MessengerClient, sep10SigningPrivateKey string, maxInvitationSMSResendAttempts int64, crashTrackerClient crashtracker.CrashTrackerClient) (*SendReceiverWalletInviteService, error) {
+func NewSendReceiverWalletInviteService(models *data.Models, messengerClient message.MessengerClient, sep10SigningPrivateKey string, maxInvitationSMSResendAttempts int64, crashTrackerClient crashtracker.CrashTrackerClient, UseExternalID bool) (*SendReceiverWalletInviteService, error) {
 	s := &SendReceiverWalletInviteService{
 		messengerClient:                messengerClient,
 		Models:                         models,
 		maxInvitationSMSResendAttempts: maxInvitationSMSResendAttempts,
 		sep10SigningPrivateKey:         sep10SigningPrivateKey,
 		crashTrackerClient:             crashTrackerClient,
+		UseExternalID:                  UseExternalID,
```
marcelosalloum, line 300: not: In Go, when a parameter starts with uppercase, it can be accessed from outside the package, so we only start things with uppercase if there's a reason for it.

```suggestion
		useExternalID:                  UseExternalID,
```

internal/services/send_receiver_wallets_invite_service.go
```diff
@@ -283,13 +290,14 @@ func (s SendReceiverWalletInviteService) shouldSendInvitationSMS(ctx context.Con
 	return true
 }
 
-func NewSendReceiverWalletInviteService(models *data.Models, messengerClient message.MessengerClient, sep10SigningPrivateKey string, maxInvitationSMSResendAttempts int64, crashTrackerClient crashtracker.CrashTrackerClient) (*SendReceiverWalletInviteService, error) {
+func NewSendReceiverWalletInviteService(models *data.Models, messengerClient message.MessengerClient, sep10SigningPrivateKey string, maxInvitationSMSResendAttempts int64, crashTrackerClient crashtracker.CrashTrackerClient, UseExternalID bool) (*SendReceiverWalletInviteService, error) {
```
marcelosalloum, line 293: Same here:
```suggestion
func NewSendReceiverWalletInviteService(models *data.Models, messengerClient message.MessengerClient, sep10SigningPrivateKey string, maxInvitationSMSResendAttempts int64, crashTrackerClient crashtracker.CrashTrackerClient, useExternalID bool) (*SendReceiverWalletInviteService, error) {
```

internal/serve/httphandler/verifiy_receiver_registration_handler.go
```diff
@@ -112,6 +112,25 @@ func (v VerifyReceiverRegistrationHandler) processReceiverVerificationPII(
 	now := time.Now()
 	truncatedPhoneNumber := utils.TruncateString(receiver.PhoneNumber, 3)
 
+	// STEP 0: if customer-id exists in sep24 claims, use it to look up reciever, hash Phone Number  in db, compare against hashed mobile_number in sep24 claims
+	if receiverRegistrationRequest.CustomerID != "" {
```
marcelosalloum, line 116: 👋  The trigger to validate the customerID should not depend on it being in the request, it should instead depend on the `use-external-id` CLI configuration.

Otherwise the externalID check will be ignored if the externalID is not provided.

This suggestion shows how to wire this configuration: https://github.com/stellar/stellar-disbursement-platform-backend/pull/256#discussion_r1563061427

internal/serve/httphandler/verifiy_receiver_registration_handler.go
```diff
@@ -112,6 +112,25 @@ func (v VerifyReceiverRegistrationHandler) processReceiverVerificationPII(
...
+		fmt.Println(receiverByExternalID.PhoneNumber)
+
+		if data.CompareVerificationValue(receiverRegistrationRequest.MobileNumberHash, receiverByExternalID.PhoneNumber) {
+			// Compare the hashed phone number with the hashed mobile number from the SEP-24 claims
+			//if hashedPhoneNumber != receiverRegistrationRequest.MobileNumberHash {
+			//err := fmt.Errorf("hashed phone number does not match the hashed mobile number from SEP-24 claims")
+			//	return &ErrorInformationNotFound{cause: err}
+			//}
+			return nil
+		}
```
marcelosalloum, line 131: 👋 👋 👋  This logic is incorrect. When you return nil here, you're skipping all the check from the following steps (STEP 1 -> 4).

What you should be doing here is returning an error if the comparison fails, similar to what's in the comment:
```suggestion
		if !data.CompareVerificationValue(receiverRegistrationRequest.MobileNumberHash, receiverByExternalID.PhoneNumber) {
			err := fmt.Errorf("hashed phone number does not match the hashed mobile number from SEP-24 claims")
			return &ErrorInformationNotFound{cause: err}
		}
```

marcelosalloum (commented):
Thanks for all the work!

Thanks for working on this, you're almost there! I marked with 👋  the changes that you need to do in order for this to work for Cassio. The changes are:
1. Change the default value of `use-external-id`, so the feature gets enabled in the PR preview
2. Wire the `use-external-id` to the SEP24 verification code. It's gonna be used in the `STEP 0` you added.
3. The `CompareVerificationValue` call in `STEP 0` is returning nil too early, making it so the user never gets marked as verified.

Plz let me know if you have any questions.

internal/data/receivers.go
```diff
@@ -31,6 +31,9 @@ type ReceiverRegistrationRequest struct {
 	VerificationValue string            `json:"verification"`
 	VerificationType  VerificationField `json:"verification_type"`
 	ReCAPTCHAToken    string            `json:"recaptcha_token"`
+	ExternalID        string            `json:"external_id"`
+	CustomerID        string            `json:"customer_id"`   //customer identifier
```
marcelosalloum, line 35: One of these will eventually be removed once we nail the customer_id vs external_id discussion.

marcelosalloum: Oh, I can see that [SEP-24 uses `customer_id`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request), so we should use the same wording.

```suggestion
	// CustomerID is the SEP-24 field that we store in our database's `receivers.external_id`.
	CustomerID        string            `json:"customer_id"`   //customer identifier
```

internal/services/send_receiver_wallets_invite_service.go
reecexlm: got it thanks

PR #331, opened 2024-06-18
[SDP-1180] Refactor CircleClient usage to use a CircleService, that automatically routes the calls by tenant

marwen-abid:
### What

Refactor CircleClient usage to use a CircleService, that automatically routes the calls by tenant

### Why

This simplifies the usage of Circle and makes it more consistent with the database connection pool usage, where resources are routed based on the tenant stored in the context.


internal/circle/client.go
```diff
@@ -23,7 +23,7 @@ const (
 
 // ClientInterface defines the interface for interacting with the Circle API.
 //
-//go:generate mockery --name=ClientInterface --case=underscore --structname=MockClient
+//go:generate mockery --name=ClientInterface --case=underscore --structname=MockClient --filename=client_mock.go --inpackage
```
marcelosalloum, line 26: This change moved the circle mocks from `.../circle/mocks` to `.../circle`, which was needed in order to allow these mocks to be used within the Circle package without falling into circular dependency.

The circular dependency was caused by many of the circle structs being needed for these mocks.

marwen-abid (commented):
LGTM! Thank you for the changes! I can't approve this because I opened this PR but feel free to approve/merge.

marcelosalloum (approved):
LGTM!

PR #338, opened 2024-06-26
[SDP-1245] Invalidate Circle Distribution Account Status upon receiving a 401/403

ziyliu:
### What
We should mark Circle distribution accounts whose API keys are not either expired (401) or lack transfer request permissions through Circle (403) as "deactivated".

### Why
Requests to SDP to process payments will not trigger any transfer request to Circle which will determinately fail.

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Configs and Secrets

#### Release

#### Deployment


stellar-multitenant/pkg/tenant/manager.go
```diff
@@ -236,6 +238,36 @@ func (m *Manager) SoftDeleteTenantByID(ctx context.Context, tenantID string) (*T
...
+		SET
+			distribution_account_status = 'PENDING_USER_ACTIVATION'
+	`
+
+	queryParams := &QueryParams{
+		Filters: excludeInactiveTenantsFilters(),
+	}
+	queryParams.Filters[FilterKeyID] = tenantID
+	queryParams.Filters[FilterKeyDistributionAccountType] = schema.DistributionAccountCircleDBVault
+	queryParams.Filters[FilterKeyDistributionAccountStatus] = schema.AccountStatusActive
```
ziyliu, line 255: I'm thinking we should explicitly return an error for when trying to deactivate an account that isn't a Circle account or active

stellar-multitenant/pkg/tenant/manager.go
```diff
@@ -236,6 +238,36 @@ func (m *Manager) SoftDeleteTenantByID(ctx context.Context, tenantID string) (*T
...
+// DeactivateTenantDistributionAccount sets a distribution account of status ACTIVE to PENDING_USER_ACTIVATION for the given tenant id,
+// and is only used in the case where the distribution account is of type CircleDBVault.
+func (m *Manager) DeactivateTenantDistributionAccount(ctx context.Context, tenantID string) error {
+	updateQuery := `
+		UPDATE tenants t
+		SET
+			distribution_account_status = 'PENDING_USER_ACTIVATION'
+	`
+
+	queryParams := &QueryParams{
```
marwen-abid, line 250: I think queryParams should be used when we have to conditionally stitch the query string together depending on optional parameters. 

In cases (like this) where the query doesn't change between calls, we should keep things explicit and just write the full query.

stellar-multitenant/pkg/tenant/manager.go
```diff
@@ -236,6 +238,36 @@ func (m *Manager) SoftDeleteTenantByID(ctx context.Context, tenantID string) (*T
...
+	queryParams.Filters[FilterKeyDistributionAccountType] = schema.DistributionAccountCircleDBVault
+	queryParams.Filters[FilterKeyDistributionAccountStatus] = schema.AccountStatusActive
+
+	query, params := m.newManagerQuery(updateQuery, queryParams)
+	query += " RETURNING *"
+
+	var t Tenant
+	if err := m.db.GetContext(ctx, &t, query, params...); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return ErrTenantDoesNotExist
```
marwen-abid, line 263: This is not accurate. There can be `ErrNoRows` when the Tenant Exists but is already deactivated, or exists and is not Circle, etc. We should still throw an error, but we can just say that the tenant was not deactivated.

internal/circle/errors.go
```diff
@@ -28,15 +32,33 @@ func (e APIError) Error() string {
 	return fmt.Sprintf("APIError: Code=%d, Message=%s, Errors=%v", e.Code, e.Message, e.Errors)
 }
 
+type TenantStatusUpdater struct {
+	tenantManager tenant.ManagerInterface
+}
+
+var invalidAPIKeyStatusCodes = []int{http.StatusUnauthorized, http.StatusForbidden}
```
marwen-abid, line 39: nit: there are other *invalid* status codes like 5xx, so we should rename this variable to something like `authErrorStatusCodes` to be more accurate.

internal/circle/errors.go
```diff
@@ -28,15 +32,33 @@ func (e APIError) Error() string {
...
+
 // parseAPIError parses the error response from Circle APIs.
 // https://developers.circle.com/circle-mint/docs/circle-apis-api-errors.
-func parseAPIError(resp *http.Response) (*APIError, error) {
+func (u TenantStatusUpdater) parseAPIError(ctx context.Context, resp *http.Response) (*APIError, error) {
```
marwen-abid, line 43: Something with this `TenantStatusUpdater` feels wrong to me. I think we're mixing concerns here. 

This `errors.go` file should be just about defining and parsing errors coming back from circle. The logic of deactivating an account should be encapsulated elsewhere (client). 

We can add a method to the client, `handleApiError(ctx, resp)` that parses the error then handles it appropriately.

internal/circle/client_test.go
```diff
@@ -290,12 +309,14 @@ func Test_Client_GetWalletByID(t *testing.T) {
 	})
 }
 
-func newClientWithMock(t *testing.T) (Client, *httpclientMocks.HttpClientMock) {
+func newClientWithMock(t *testing.T) (Client, *httpclientMocks.HttpClientMock, *tenant.TenantManagerMock) {
```
marcelosalloum, line 312: 💡 now we have multiple mocks, so we should use the plural here.

```suggestion
func newClientWithMocks(t *testing.T) (Client, *httpclientMocks.HttpClientMock, *tenant.TenantManagerMock) {
```

internal/circle/client_test.go
```diff
@@ -290,12 +309,14 @@ func Test_Client_GetWalletByID(t *testing.T) {
...
 	httpClientMock := httpclientMocks.NewHttpClientMock(t)
+	tntManagerMock := &tenant.TenantManagerMock{}
```
marcelosalloum, line 314: 💡  If you use `tenant.NewTenantManagerMock(t)`, the AssertExpectations(t) should be called automatically.

internal/circle/service.go
```diff
@@ -103,7 +111,7 @@ func (s *Service) getClient(ctx context.Context) (ClientInterface, error) {
 	if err != nil {
 		return nil, fmt.Errorf("retrieving decrypted Circle API key: %w", err)
 	}
-	return s.ClientFactory(s.NetworkType, apiKey), nil
+	return s.ClientFactory(s.NetworkType, apiKey, s.TenantManager), nil
```
marcelosalloum, line 114: 🤔  Now that I'm reading this class again, I gotta say it's not 100% clear that the client being instantiated here uses the apiKey from the tenant that is found in the context.

I know this is not something you introduced, but could you improve the naming of the function to make it clearer?Maybe `getClientForTenantInContext` or `getClientForContextTenant`, or something like that?

internal/circle/service_test.go
marcelosalloum, line 1: 👋  Can you add a test to make sure the tenant updater is being called to handle 401/403 in the circle service?

stellar-multitenant/pkg/tenant/manager.go
marcelosalloum: I think that'd be simpler as well.

stellar-multitenant/pkg/tenant/manager.go
marcelosalloum: Good point, IMO we don't need to throw Not sure we need to throw this error when the status is not Active.

marcelosalloum (commented):
I echo Marwen's feedback and I left a few comments as well. The most important one is marked with a 👋

stellar-multitenant/pkg/tenant/manager.go
marwen-abid: We don't need to throw an error, but if I follow the logic correctly, we should never get in the case where we attempt deactivating an account that's already deactivated. 

I'm fine with not throwing an error though. It makes the operation idempotent.

ziyliu: so to summarize
- we should hard code all the filters in the query params in the sql statement itself
- don’t need to return an error indicating whether a tenant was updated at all, making the operation just idempotent

PR #368, opened 2024-07-26
[SDP-1247] new API role for bypassing reCAPTCHA

ziyliu:
### What
Completed:
- add role `APIUserRole` that could be concatenated with existing user roles to override reCAPTCHA validation
- local testing for all three endpoints `/login`, `/mfa`, and `/forgot-password` that would usually require reCAPTCHA validation normally
- fixes to existing unit tests and new tests for data access methods
- refactoring fixed unit tests to be less repetitive

##Tests
The following calls were made referring to users with the `api` role and without the recaptcha token in the request body

`/login`
request
```
{
    "email": "owner@bluecorp.local",
    "password": <REDACTED>
}
```
response
```
{
    "token": "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjp7ImlkIjoiOWI3OTI4MWEtY2E1OS00MDZjLWIwNmMtYTE3NWNhYmQ0ZDA5IiwiZmlyc3RfbmFtZSI6ImpvaG4iLCJsYXN0X25hbWUiOiJkb2UiLCJlbWFpbCI6Im93bmVyQGJsdWVjb3JwLmxvY2FsIiwiaXNfYWN0aXZlIjpmYWxzZSwicm9sZXMiOlsib3duZXIiXX0sInRlbmFudF9pZCI6ImFhYzFlY2ZjLTQzZDItNDlkYy1iMzE3LTJjOTA4MzRmMjM5YiIsImV4cCI6MTcyMTk0MTk2NH0.O_dDRFXGV9tWhE8gIuxSPn8SUMsmJphJkAyVYEQ-5CBMiZ3egKKMEe_BHZLfc3bp3TVhMMB4Dc4JjGNs-Ez2nw"
}
```

```
26e4ca25-94b7-40a0-97e4-31c4d7271826 | 9b79281a-ca59-406c-b06c-a175cabd4d09 |  123  | 2024-07-26 00:31:24.611207+00 | 2024-07-26 09:40:56.524779+00 | 2024-07-28 00:31:24.611207+00 |  
-- | -- | -- | -- | -- | -- | --
```

`/mfa`

request
```
{
    "mfa_code": "123",
    "remember_me": false
}
```

response
```
{
    "token": "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjp7ImlkIjoiOWI3OTI4MWEtY2E1OS00MDZjLWIwNmMtYTE3NWNhYmQ0ZDA5IiwiZmlyc3RfbmFtZSI6ImpvaG4iLCJsYXN0X25hbWUiOiJkb2UiLCJlbWFpbCI6Im93bmVyQGJsdWVjb3JwLmxvY2FsIiwiaXNfYWN0aXZlIjpmYWxzZSwicm9sZXMiOlsib3duZXIiXX0sInRlbmFudF9pZCI6ImFhYzFlY2ZjLTQzZDItNDlkYy1iMzE3LTJjOTA4MzRmMjM5YiIsImV4cCI6MTcyMTk1ODU0Nn0.aMqfzwTTt8tPTgbyyai5FurrWwol9LDnrI6KexnSPGshghGdtUlZCy_ssfGug96qS5D8RG3_qNnFW0-0lzGmpw"
}
```

`/forgot-password`
request
```
{
    "email": "owner@bluecorp.local"
}
```
response
```
{
    "message": "Password reset requested. If the email is registered, you'll receive a reset link shortly. Check your inbox and spam folders."
}
```

### Why

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Configs and Secrets

#### Release

* [ ] This is not a breaking change.

#### Deployment

### DIFF
--- a/internal/serve/httphandler/forgot_password_handler.go
+++ b/internal/serve/httphandler/forgot_password_handler.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/support/render/httpjson"
@@ -59,7 +60,28 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if !h.ReCAPTCHADisabled {
+	// validate request
+	v := validators.NewValidator()
+
+	v.Check(forgotPasswordRequest.Email != "", "email", "email is required")
+
+	if v.HasErrors() {
+		httperror.BadRequest("request invalid", err, v.Errors).Render(w)
+		return
+	}
+
+	user, err := h.AuthManager.GetUserByEmail(ctx, forgotPasswordRequest.Email)
+	if err != nil {
+		if errors.Is(err, auth.ErrUserNotFound) {
+			log.Ctx(ctx).Errorf("error in forgot password handler, email not found: %s", forgotPasswordRequest.Email)
+		} else {
+			httperror.InternalError(ctx, "getting user by email", err, nil).Render(w)
+			return
+		}
+	}
+	canBypassReCAPTCHA := slices.Contains(user.Roles, data.APIOwnerUserRole.String())
+
+	if !h.ReCAPTCHADisabled && !canBypassReCAPTCHA {
 		// validating reCAPTCHA Token
 		isValid, recaptchaErr := h.ReCAPTCHAValidator.IsTokenValid(ctx, forgotPasswordRequest.ReCAPTCHAToken)
 		if recaptchaErr != nil {
@@ -72,16 +94,6 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 			httperror.BadRequest("reCAPTCHA token invalid", nil, nil).Render(w)
 			return
 		}
-	}
-
-	// validate request
-	v := validators.NewValidator()
-
-	v.Check(forgotPasswordRequest.Email != "", "email", "email is required")
-
-	if v.HasErrors() {
-		httperror.BadRequest("request invalid", err, v.Errors).Render(w)
-		return
 	}
 
 	resetToken, err := h.AuthManager.ForgotPassword(ctx, forgotPasswordRequest.Email)
--- a/internal/serve/httphandler/forgot_password_handler.go
+++ b/internal/serve/httphandler/forgot_password_handler.go
@@ -56,7 +56,7 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	var forgotPasswordRequest ForgotPasswordRequest
 	err = json.NewDecoder(r.Body).Decode(&forgotPasswordRequest)
 	if err != nil {
-		httperror.BadRequest("invalid request body", err, nil).Render(w)
+		httperror.BadRequest("Invalid request body", err, nil).Render(w)
 		return
 	}
 
@@ -66,7 +66,7 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	v.Check(forgotPasswordRequest.Email != "", "email", "email is required")
 
 	if v.HasErrors() {
-		httperror.BadRequest("request invalid", err, v.Errors).Render(w)
+		httperror.BadRequest("Request invalid", err, v.Errors).Render(w)
 		return
 	}
 
@@ -75,13 +75,17 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		if errors.Is(err, auth.ErrUserNotFound) {
 			// If we don't find the user by email, we just return an ok response
 			// to prevent malicious client from searching accounts in the system
-			log.Ctx(ctx).Errorf("email in request not found: %s", forgotPasswordRequest.Email)
+			log.Ctx(ctx).Errorf("Email in request not found: %s", forgotPasswordRequest.Email)
 		} else {
-			httperror.InternalError(ctx, "getting user by email", err, nil).Render(w)
+			httperror.InternalError(ctx, "Getting user by email", err, nil).Render(w)
 			return
 		}
 	}
-	userRoleCanBypassReCAPTCHA := slices.Contains(user.Roles, data.APIUserRole.String())
+
+	userRoleCanBypassReCAPTCHA := false
+	if user != nil {
+		userRoleCanBypassReCAPTCHA = slices.Contains(user.Roles, data.APIUserRole.String())
+	}
 
 	if !h.ReCAPTCHADisabled && !userRoleCanBypassReCAPTCHA {
 		// validating reCAPTCHA Token
@@ -103,9 +107,9 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	// to prevent malicious client from searching accounts in the system
 	if err != nil {
 		if errors.Is(err, auth.ErrUserNotFound) {
-			log.Ctx(ctx).Errorf("error in forgot password handler, email not found: %s", forgotPasswordRequest.Email)
+			log.Ctx(ctx).Errorf("Email not found in forgot password handler: %s", forgotPasswordRequest.Email)
 		} else if errors.Is(err, auth.ErrUserHasValidToken) {
-			log.Ctx(ctx).Errorf("error in forgot password handler, user has a valid token")
+			log.Ctx(ctx).Errorf("User has a valid token in forgot password handler")
 		} else {
 			httperror.InternalError(ctx, "", err, nil).Render(w)
 			return
@@ -115,14 +119,14 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	if err == nil {
 		organization, err := h.Models.Organizations.Get(ctx)
 		if err != nil {
-			err = fmt.Errorf("error getting organization data: %w", err)
+			err = fmt.Errorf("getting organization data: %w", err)
 			httperror.InternalError(ctx, "", err, nil).Render(w)
 			return
 		}
 
 		resetPasswordLink, err := url.JoinPath(*tnt.SDPUIBaseURL, "reset-password")
 		if err != nil {
-			err = fmt.Errorf("error getting reset password link: %w", err)
+			err = fmt.Errorf("getting reset password link: %w", err)
 			log.Ctx(ctx).Error(err)
 			httperror.InternalError(ctx, "", err, nil).Render(w)
 			return
@@ -135,7 +139,7 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		}
 		messageContent, err := htmltemplate.ExecuteHTMLTemplateForForgotPasswordMessage(forgotPasswordData)
 		if err != nil {
-			err = fmt.Errorf("error executing forgot password message template: %w", err)
+			err = fmt.Errorf("executing forgot password message template: %w", err)
 			httperror.InternalError(ctx, "", err, nil).Render(w)
 			return
 		}
@@ -147,7 +151,7 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		}
 		err = h.MessengerClient.SendMessage(msg)
 		if err != nil {
-			err = fmt.Errorf("error sending forgot password email for email %s: %w", forgotPasswordRequest.Email, err)
+			err = fmt.Errorf("sending forgot password email for email %s: %w", forgotPasswordRequest.Email, err)
 			httperror.InternalError(ctx, "", err, nil).Render(w)
 			return
 		}
--- a/internal/serve/httphandler/forgot_password_handler.go
+++ b/internal/serve/httphandler/forgot_password_handler.go
@@ -59,6 +59,21 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	if !h.ReCAPTCHADisabled {
+		// validating reCAPTCHA Token
+		isValid, recaptchaErr := h.ReCAPTCHAValidator.IsTokenValid(ctx, forgotPasswordRequest.ReCAPTCHAToken)
+		if recaptchaErr != nil {
+			httperror.InternalError(ctx, "Cannot validate reCAPTCHA token", recaptchaErr, nil).Render(w)
+			return
+		}
+
+		if !isValid {
+			log.Ctx(ctx).Errorf("reCAPTCHA token is invalid for request with email %s", utils.TruncateString(forgotPasswordRequest.Email, 3))
+			httperror.BadRequest("reCAPTCHA token invalid", nil, nil).Render(w)
+			return
+		}
+	}
+
 	// validate request
 	v := validators.NewValidator()
 
@@ -79,28 +94,6 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 			log.Ctx(ctx).Errorf("error in forgot password handler, user has a valid token")
 		} else {
 			httperror.InternalError(ctx, "", err, nil).Render(w)
-			return
-		}
-	}
-
-	roleCanBypassReCAPTCHA, err := validators.UserRoleCanBypassReCAPTCHA(ctx, h.AuthManager, resetToken)
-	if err != nil {
-		log.Ctx(ctx).Errorf("error checking if user role in token can bypass ReCAPTCHA: %s", err)
-		httperror.InternalError(ctx, "", err, nil).Render(w)
-		return
-	}
-
-	if !h.ReCAPTCHADisabled && !roleCanBypassReCAPTCHA {
-		// validating reCAPTCHA Token
-		isValid, recaptchaErr := h.ReCAPTCHAValidator.IsTokenValid(ctx, forgotPasswordRequest.ReCAPTCHAToken)
-		if recaptchaErr != nil {
-			httperror.InternalError(ctx, "Cannot validate reCAPTCHA token", recaptchaErr, nil).Render(w)
-			return
-		}
-
-		if !isValid {
-			log.Ctx(ctx).Errorf("reCAPTCHA token is invalid for request with email %s", utils.TruncateString(forgotPasswordRequest.Email, 3))
-			httperror.BadRequest("reCAPTCHA token invalid", nil, nil).Render(w)
 			return
 		}
 	}
--- a/internal/serve/httphandler/forgot_password_handler.go
+++ b/internal/serve/httphandler/forgot_password_handler.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"strings"
 
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/support/render/httpjson"
@@ -72,13 +73,11 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 
 	user, err := h.AuthManager.GetUserByEmail(ctx, forgotPasswordRequest.Email)
 	if err != nil {
+		// If user cannot be found, we will default to user requiring to validate reCAPTCHA.
 		if errors.Is(err, auth.ErrUserNotFound) {
 			// If we don't find the user by email, we just return an ok response
 			// to prevent malicious client from searching accounts in the system
-			log.Ctx(ctx).Errorf("Email in request not found: %s", forgotPasswordRequest.Email)
-		} else {
-			httperror.InternalError(ctx, "Getting user by email", err, nil).Render(w)
-			return
+			log.Ctx(ctx).Errorf("Email in request not found: %s", utils.TruncateString(forgotPasswordRequest.Email, 3))
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -57,51 +61,82 @@ func Test_ForgotPasswordHandler(t *testing.T) {
...
+		usersRoles := make([][]string, 2)
+		// role cannot bypass reCAPTCHA
+		usersRoles[0] = []string{data.OwnerUserRole.String()}
+		// role can bypass reCAPTCHA
+		usersRoles[1] = []string{data.OwnerUserRole.String(), data.APIUserRole.String()}
+		for _, userRoles := range usersRoles {
+			requestBody := fmt.Sprintf(`
+				{ 
+					"email": "%s"
+				}`, user.Email)
```
marcelosalloum, line 97: 💡  This setup is confusing to me. Can you create two users, one with the API role and one that doesn't have it? That would make things easier to read and also make the code easier to maintain.

internal/serve/httphandler/login_handler.go
```diff
@@ -101,6 +87,23 @@ func (h LoginHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+			return
+		}
+
+		if !isValid {
+			log.Ctx(ctx).Errorf("reCAPTCHA token is invalid for request with email %s", utils.TruncateString(reqBody.Email, 3))
+			httperror.BadRequest("reCAPTCHA token invalid", nil, nil).Render(rw)
+			return
+		}
+	}
+
```
marcelosalloum, line 106: 🚨 the purpose of having recaptcha validated before the authentication is to prevent bots from brute forcing their way in.

By moving this recaptcha after the authentication call, you're removing these safety guardrails from this endpoint.

internal/serve/httphandler/mfa_handler.go
```diff
@@ -81,11 +67,29 @@ func (h MFAHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+
+		if !isValid {
+			log.Ctx(ctx).Errorf("reCAPTCHA token is invalid for request with email")
+			httperror.BadRequest("reCAPTCHA token invalid", nil, nil).Render(rw)
+			return
+		}
+	}
+
+	log.Ctx(ctx).Infof("[UserLogin] - Logged in user with account ID %s", user.ID)
```
marcelosalloum, line 93: 🚨 Same here. By moving the recaptcha part past the authentication calls, you;re defeating the purpose of the recaptcha.

internal/serve/httphandler/user_handler_test.go
marcelosalloum, line 1: Can you add tests to ensure that patching the user roles with {APIRole + SomeOtherRole} succeeds?

marcelosalloum (changes requested):
The way the recaptcha was altered in some endpoints poses a security issue. I've marked those comments with 🚨. I didn't finish reviewing the whole PR as those issues caught my attention as highest priority ones, I'll do another review after they're fixed.

There are also a couple of other suggestions I marked in the code.

internal/serve/httphandler/login_handler.go
ziyliu: Addressed [abeb515](https://github.com/stellar/stellar-disbursement-platform-backend/pull/368/commits/abeb515c89402d855ecb6723f3e4f12c94cdd3c7)

internal/serve/httphandler/mfa_handler.go
ziyliu: Addressed [abeb515](https://github.com/stellar/stellar-disbursement-platform-backend/pull/368/commits/abeb515c89402d855ecb6723f3e4f12c94cdd3c7)

internal/serve/httphandler/user_handler_test.go
ziyliu: yes, great idea!

internal/serve/httphandler/forgot_password_handler.go
```diff
@@ -55,11 +56,38 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+	v := validators.NewValidator()
+
+	v.Check(forgotPasswordRequest.Email != "", "email", "email is required")
+
+	if v.HasErrors() {
+		httperror.BadRequest("Request invalid", err, v.Errors).Render(w)
+		return
+	}
+
+	user, err := h.AuthManager.GetUserByEmail(ctx, forgotPasswordRequest.Email)
```
marcelosalloum, line 73: 💡  For a better UX, it'd be good to sanitize the email field by stripping any extra space and lowercasing it.

internal/serve/httphandler/forgot_password_handler.go
```diff
@@ -55,11 +56,38 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		if errors.Is(err, auth.ErrUserNotFound) {
+			// If we don't find the user by email, we just return an ok response
+			// to prevent malicious client from searching accounts in the system
+			log.Ctx(ctx).Errorf("Email in request not found: %s", forgotPasswordRequest.Email)
```
marcelosalloum, line 78: 👋  Leaking PII into the logs is considered a security problem. If a bad actor get access to the logs, they'd be able to recover emails from known users, and use that to narrow down their targets.

Whenever we log something like this, we usually truncate it with the function `utils.TruncateString`.

internal/serve/httphandler/forgot_password_handler.go
```diff
@@ -74,24 +102,14 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
-	// if we don't find the user by email, we just return an ok response
+	// If we don't find the user by email, we just return an ok response
 	// to prevent malicious client from searching accounts in the system
 	if err != nil {
 		if errors.Is(err, auth.ErrUserNotFound) {
-			log.Ctx(ctx).Errorf("error in forgot password handler, email not found: %s", forgotPasswordRequest.Email)
+			log.Ctx(ctx).Errorf("Email not found in forgot password handler: %s", forgotPasswordRequest.Email)
 		} else if errors.Is(err, auth.ErrUserHasValidToken) {
-			log.Ctx(ctx).Errorf("error in forgot password handler, user has a valid token")
+			log.Ctx(ctx).Errorf("User has a valid token in forgot password handler")
```
marcelosalloum, line 112: 🤔 It's not clear what that means. After looking the code, it seems the forgot password will generate a maximum of one token per 20 minutes. meaning the user would need to wait for 20min before they can generate a new token.

💡  Can you improve the log message to explain that, so an admin can debug the problem without knowledge of the codebase?

internal/serve/httphandler/forgot_password_handler.go
```diff
@@ -133,7 +151,7 @@ func (h ForgotPasswordHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		}
 		err = h.MessengerClient.SendMessage(msg)
 		if err != nil {
-			err = fmt.Errorf("error sending forgot password email for email %s: %w", forgotPasswordRequest.Email, err)
+			err = fmt.Errorf("sending forgot password email for email %s: %w", forgotPasswordRequest.Email, err)
```
marcelosalloum, line 154: Thank you for improving these error messages.

internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -57,51 +72,77 @@ func Test_ForgotPasswordHandler(t *testing.T) {
...
-			Return(true, nil).
-			Once()
-
-		resetPasswordLink, err := urllib.JoinPath(uiBaseURL, "reset-password")
-		require.NoError(t, err)
+		usersRoles := make([][]string, 2)
+		// role cannot bypass reCAPTCHA
+		usersRoles[0] = []string{data.OwnerUserRole.String()}
+		// role can bypass reCAPTCHA
+		usersRoles[1] = []string{data.OwnerUserRole.String(), data.APIUserRole.String()}
```
marcelosalloum, line 93: 💡 This could be a little bit more readable IMO:
```suggestion
		usersRoles := [][]string{
			{data.OwnerUserRole.String()},                            // cannot bypass reCAPTCHA
			{data.OwnerUserRole.String(), data.APIUserRole.String()}, // API role can bypass reCAPTCHA
		}
```

internal/serve/httphandler/forgot_password_handler_test.go
ziyliu: I fixed this to include a user with each of the list of user roles! let me know if it's still unclear to you

internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -57,51 +72,77 @@ func Test_ForgotPasswordHandler(t *testing.T) {
...
+				{ 
+					"email": "%s"
+				}`, targetUser.Email)
+
+			authenticatorMock.
+				On("ForgotPassword", mock.Anything, email).
+				Return("resetToken", nil).
+				Once()
+			UserByEmailSetup(authenticatorMock, roleManagerMock, targetUser, userRoles)
+			if !slices.Contains(userRoles, data.APIUserRole.String()) {
```
marcelosalloum, line 110: 💡  We should highlight that this piece of code is where the condition of `usersRoles` is used:
```suggestion
			if !slices.Contains(userRoles, data.APIUserRole.String()) {  // <-------- bypasses recaptcha when APIUserRole is present.
```

internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -57,51 +72,77 @@ func Test_ForgotPasswordHandler(t *testing.T) {
 		SDPUIBaseURL: &uiBaseURL,
 	}
 	ctx := tenant.SaveTenantInContext(context.Background(), &tnt)
+	const email = "valid@email.com"
+	defaultUserRoles := []string{data.OwnerUserRole.String()}
```
marcelosalloum, line 76: When I read `defaultUserRoles` throughout this file, I need to come back often to its definition to understand which role is the default. We should use a name that conveys the meaning of the value. What do you think about calling it `ownerUserRoleSlice`?

Or maybe just passing the const data.OwnerUserRole?

If your goal here is to avoid instantiating the []string slice every time, you could use [variadic functions](https://gobyexample.com/variadic-functions).

internal/serve/httphandler/login_handler.go
```diff
@@ -68,11 +69,28 @@ func (h LoginHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if !h.ReCAPTCHADisabled {
+	user, err := h.AuthManager.GetUserByEmail(ctx, reqBody.Email)
```
marcelosalloum, line 72: 💡  Same suggestion here about sanitizing the email

internal/serve/httphandler/login_handler.go
```diff
@@ -68,11 +69,28 @@ func (h LoginHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		if errors.Is(err, auth.ErrUserNotFound) {
+			// If we don't find the user by email, we just return an ok response
+			// to prevent malicious client from searching accounts in the system
+			log.Ctx(ctx).Errorf("Email in request not found: %s", reqBody.Email)
```
marcelosalloum, line 77: 👋 same here about truncating the PII

internal/serve/httphandler/login_handler_test.go
```diff
@@ -24,6 +26,37 @@ import (
 	"github.com/stellar/stellar-disbursement-platform-backend/stellar-auth/pkg/auth"
 )
 
+func authenticateSetup(
+	authenticatorMock *auth.AuthenticatorMock,
+	roleManagerMock *auth.RoleManagerMock,
+	jwtManagerMock *auth.JWTManagerMock,
+	user *auth.User, password, userToken string,
+) {
+	authenticatorMock.
```
marcelosalloum, line 35: 💡 
```suggestion
) {
	t.Helper()
	authenticatorMock.
```

stellar-auth/pkg/auth/authenticator.go
```diff
@@ -522,6 +523,32 @@ func (a *defaultAuthenticator) GetUsers(ctx context.Context, userIDs []string) (
...
+		}
+		return nil, fmt.Errorf("querying user email %s: %w", email, err)
+	}
+
+	return &User{
+		ID:    u.ID,
+		Email: email,
+	}, nil
+}
+
```
marcelosalloum, line 551: 🤔 💡   Why don't you fetch the roles as well? I mean, 100% of the times where you fetch the user by email, you will need to look at the user roles as well, right? Aren;'t the roles on this very same table?

marcelosalloum: TBH, `GetUser` could be returning these roles as well, but that doesn't need to be changed in this PR.

internal/serve/httphandler/mfa_handler.go
```diff
@@ -43,8 +44,36 @@ func (h MFAHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	user, err := h.AuthManager.GetUserByMFA(ctx, deviceID)
+	if err != nil {
+		if errors.Is(err, auth.ErrMFADeviceNotFound) {
+			// If we don't find the user by device id, we just return an ok response
+			// to prevent malicious client from searching for the device in the system
+			log.Ctx(ctx).Errorf("Device id in request not found: %s", deviceID)
+		}
+		httperror.InternalError(ctx, "Cannot get user by MFA", err, nil).Render(rw)
+		return
+	}
```
marcelosalloum, line 68: 🚨 having this piece of code before the recaptcha code causes the same security flaw as I raised last time on this same PR.

MFA is just 6 digits, it can be easily brute-forced if this return message is not placed **after** the recaptcha

internal/serve/httphandler/mfa_handler.go
```diff
@@ -43,8 +44,36 @@ func (h MFAHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+
+	if reqBody.MFACode == "" {
+		extras := map[string]interface{}{"mfa_code": "MFA Code is required"}
+		httperror.BadRequest("Request invalid", nil, extras).Render(rw)
+		return
+	}
+
+	user, err := h.AuthManager.GetUserByMFA(ctx, deviceID)
```
marcelosalloum, line 59: 👋  Why is this method called `GetUserByMFA` if what you're passing is the deviceID and not the MFA?

stellar-auth/pkg/auth/authenticator_test.go
```diff
@@ -1039,3 +1039,36 @@ func Test_DefaultAuthenticator_UpdatePassword(t *testing.T) {
...
+			On("Encrypt", ctx, mock.AnythingOfType("string")).
+			Return("encryptedPassword", nil)
+
+		randUser := CreateRandomAuthUserFixture(t, ctx, dbConnectionPool, passwordEncrypterMock, false, "role1")
+
+		u, err := authenticator.GetUserByEmail(ctx, randUser.Email)
+		require.NoError(t, err)
+
+		assert.Equal(t, randUser.ID, u.ID)
+		assert.Equal(t, randUser.Email, u.Email)
```
marcelosalloum, line 1072: In the assertion, it's better to compare the whole user object, so you can make sure that the getter is retrieving all the fields you're expecting.

stellar-auth/pkg/auth/mfa_manager.go
```diff
@@ -147,6 +149,31 @@ func (m *defaultMFAManager) ForgetDevice(ctx context.Context, deviceID, userID s
 	return nil
 }
 
+func (m *defaultMFAManager) GetAuthUserID(ctx context.Context, deviceID string) (string, error) {
```
marcelosalloum, line 152: 👀 `auth` is just a prefix used in the database for auth-related tables. The correct naming here according with the rest of the code conventions would be `GetUserID` or `GetUserIDByDevideID`.

stellar-auth/pkg/auth/mfa_manager.go
```diff
@@ -35,6 +36,7 @@ const (
 var (
 	ErrMFACodeInvalid         = errors.New("MFA code is invalid")
 	ErrMFANoCodeForUserDevice = errors.New("no MFA code for user and device")
+	ErrMFADeviceNotFound      = errors.New("MFA device not found")
```
marcelosalloum, line 39: 💡 
```suggestion
	ErrMFADeviceIDNotFound      = errors.New("MFA device ID not found")
```

stellar-auth/pkg/auth/mfa_manager.go
```diff
@@ -147,6 +149,31 @@ func (m *defaultMFAManager) ForgetDevice(ctx context.Context, deviceID, userID s
...
+		WHERE 
+			device_id = $1
+	`
+	var userID string
+	err := m.dbConnectionPool.GetContext(ctx, &userID, query, deviceID)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return "", ErrMFADeviceNotFound
+		}
+		return "", fmt.Errorf("error fetching user ID for device ID %s: %w", deviceID, err)
```
marcelosalloum, line 171: Can you update this error message so it doesn't start with `error`?

internal/serve/httphandler/login_handler.go
```diff
@@ -68,11 +69,28 @@ func (h LoginHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		} else {
+			httperror.InternalError(ctx, "Getting user by email", err, nil).Render(rw)
```
marcelosalloum, line 79: 🚨  We shouldn't return any hints of what went wrong during the authentication, especially before recaptcha, at the risk of providing too much information to a bad actor that's trying to brute force the system.

```suggestion
			httperror.InternalError(ctx, "", err, nil).Render(rw)
```

marcelosalloum (changes requested):
I still found similar issues as the previous review. I've reviewed most of the code now and shared more feedback below.

stellar-auth/pkg/auth/authenticator_test.go
ziyliu: `randUser` creates a full fledged user fixture but `GetUserByEmail` only retrieves the fields I want to compare thus my choice to compare only the relevant fields here!

internal/serve/httphandler/forgot_password_handler.go
ziyliu: done!

internal/serve/httphandler/forgot_password_handler_test.go
ziyliu: good idea, I think the naming I introduced is not too intuitive for someone touching this test file for the first time

internal/serve/httphandler/mfa_handler.go
ziyliu: The naming for this method is misleading, so I've renamed it to `GetUserByDeviceID` per your suggestion below! I've also truncated the device ID in the log the same way we are doing for emails when searching for user by email in the other handlers

internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -57,51 +61,79 @@ func Test_ForgotPasswordHandler(t *testing.T) {
...
+			{data.OwnerUserRole.String()},                            // API roles cannot bypass reCAPTCHA
+			{data.OwnerUserRole.String(), data.APIUserRole.String()}, // API roles can bypass reCAPTCHA
 		}
-		messengerClientMock.
-			On("SendMessage", msg).
-			Return(nil).
-			Once()
+		for _, userRoles := range usersRoles {
+			targetUser := &auth.User{
+				ID:    "user-ID",
```
marcelosalloum, line 84: A few lines above you used `userID` and here you have `user-ID`. Is that a typo? Is that field being used for anything? Should we add a test for it?

internal/serve/httphandler/forgot_password_handler_test.go
```diff
@@ -380,6 +402,35 @@ func Test_ForgotPasswordHandler(t *testing.T) {
 		assert.JSONEq(t, wantsBody, string(respBody))
 	})
 
+	t.Run("Should return http status 400 when attempting to validate reCAPTCHA token after failing to find user by email", func(t *testing.T) {
+		requestBody := `
+		{ 
+			"email": "valid@email.com"
```
marcelosalloum, line 408: Would this be a better naming? WDYT?
```suggestion
			"email": "not-found@email.com"
```

internal/serve/httphandler/login_handler.go
```diff
@@ -68,11 +70,26 @@ func (h LoginHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+			// If we don't find the user by email, we just return an ok response
+			// to prevent malicious client from searching accounts in the system
+			log.Ctx(ctx).Errorf("Email in request not found: %s", utils.TruncateString(reqBody.Email, 3))
+		}
+	}
+
+	userRoleCanBypassReCAPTCHA := false
+	if user != nil {
+		userRoleCanBypassReCAPTCHA = slices.Contains(user.Roles, data.APIUserRole.String())
+	}
```
marcelosalloum, line 86: Same here about simplifying this statement.

internal/serve/httphandler/login_handler_test.go
```diff
@@ -71,6 +90,17 @@ func Test_LoginHandler(t *testing.T) {
 	}
 
 	const url = "/login"
+	const email = "testuser@email.com"
+	const password = "pass1234"
+	const defaultReCAPTCHAToken = "XyZ"
+	defaultReqBody := fmt.Sprintf(
+		`{"email": "%s", "password": "%s", "recaptcha_token": "%s"}`, email, password, defaultReCAPTCHAToken)
```
marcelosalloum, line 97: I think the naming here can be improved for better clarity. I mentioned this in a previous review on this same PR, but what does default mean vs non-default? If this is meant to be a login request body with or without recaptcha, you should improve make sure the const tame conveys that meaning.

internal/serve/httphandler/login_handler_test.go
```diff
@@ -24,6 +26,23 @@ import (
...
+	authenticatorMock.
+		On("ValidateCredentials", mock.Anything, user.Email, password).
+		Return(user, nil).
+		Once()
+	jwtManagerMock.
+		On("GenerateToken", mock.Anything, user, mock.AnythingOfType("time.Time")).
+		Return(userToken, nil).
+		Once()
+}
+
```
marcelosalloum, line 45: Can you be more specific on this method name? Maybe `prepareMocksForLogin` would be more accurate?

internal/serve/httphandler/mfa_handler.go
```diff
@@ -43,8 +46,35 @@ func (h MFAHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	if reqBody.MFACode == "" {
+		extras := map[string]interface{}{"mfa_code": "MFA Code is required"}
+		httperror.BadRequest("Request invalid", nil, extras).Render(rw)
+		return
+	}
+
+	user, err := h.AuthManager.GetUserByDeviceID(ctx, strings.TrimSpace(deviceID))
```
marcelosalloum, line 61: 🚨 It doesn't look like the implementation of `GetUserByDeviceID` is returning the user roles.

marcelosalloum (commented):
Thank you for working on that.

I still found a few issues. Comments below ⬇️

PR #512, opened 2025-01-24
[SDP-1523] allow patching already confirmed verification fields

marcelosalloum:
### What

Allow patching already confirmed verification fields

### Why

Some receivers got locked out from a partner's system and this seems to be the most reliable fix.

### Pending work

@marwen-abid added an audit table in #513 that will be used to keep track of such changes.


marwen-abid (approved):
LGTM

PR #560, opened 2025-02-26
SDP-1548 - Setting up the basics for serving SEP-24 through a React app

marwen-abid:
### Backend API 
- [x] change SEP24 endpoints to return `error_code` when there is an error. This will be used for translations. – #597 
- [x] look into returning the recaptcha key as part of the `/init` payload.  #592 
- [x] cleanup legacy flow – #608

### Frontend 
- [x] use the `error_code` for displaying a localized error to the user. – #598
- [x] use recaptcha key from the `/init` endpoint instead of `.env` file. – #593 
- [x] decom `.env` file – #593 
- [x] change the behaviour of the `finish` button to close the pop-up. #595 
- [x] Handle usecase where the user has already been registered `is_registered:true` – #594
- [x] Support recaptcha disabled flow – #599   

### CI
- [x] Add job to ensure that `yarn build` was executed and the server is serving the latest version of SEP-24 flow. – #600


internal/serve/sep24frontend/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+The application will be available at the path configured in your Go server, usually at `/wallet-registration-fe/`.
+
+## Important Note
+
+**After making any changes to the frontend application:**
+
+1. Rebuild the application with `yarn build`
+2. Commit the updated `dist` folder to the repository
```
marcelosalloum, line 48: We can add a `script` entry to `package.json` that does git add `(...)/dist`

marcelosalloum (commented):
Looks good so far 👍

marcelosalloum (approved):

PR #561, opened 2025-02-27
[SDP-1554] Refactor SEP24 Endpoints

marwen-abid:


internal/serve/httphandler/receiver_registration.go
```diff
@@ -74,34 +56,22 @@ func (h ReceiverRegistrationHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
 	if organization.PrivacyPolicyLink != nil {
 		privacyPolicyLink = *organization.PrivacyPolicyLink
 	}
-	tmplData := ReceiverRegistrationData{
-		StellarAccount:    sep24Claims.SEP10StellarAccount(),
-		JWTToken:          token,
```
marcelosalloum, line 79: The reason to extract the token from the initial request was to inject it in the frontend when rendering the first template. The FE would then use it to authenticate the subsequent steps in the SEP-24 registration flow.

By removing it, how do you plan to authenticate the FE calls to `POST (...)/otp` and `POST (...)/verification`?

marcelosalloum (approved):
Approving the PR based on the offline conversation we had where @marwen-abid said:

_"My assumption is that we'll create a sessionStorage on the first call to `/wallet-registration/start?token=xxx` (same as when you sign in) and resend that same token on subsequent calls"_

internal/serve/httphandler/receiver_registration.go
marwen-abid: My assumption is that we'll create a sessionStorage on the first call to /wallet-registration/start?token=xxx (same as when you sign in) and resend that same token on subsequent calls

PR #606, opened 2025-03-31
SDP-1584: Fix csf bug

marcelosalloum:
### What

Fix CSF bug that would prevent local images to be loaded by the backend

### Why

Screenshot 2025-03-25 at 12 29 37 PM

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/middleware/middleware.go
+++ b/internal/serve/middleware/middleware.go
@@ -275,6 +275,7 @@ func CSPMiddleware() func(http.Handler) http.Handler {
 				{"frame-ancestors", []string{selfSrc}},
 
 				{"form-action", []string{selfSrc}},
+				{"img-src", []string{selfSrc, "data:", "https:"}},
 			}
 			cspStr := ""
 			for _, item := range cspItems {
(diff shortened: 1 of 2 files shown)


internal/serve/middleware/middleware.go
```diff
@@ -275,6 +275,7 @@ func CSPMiddleware() func(http.Handler) http.Handler {
 				{"frame-ancestors", []string{selfSrc}},
 
 				{"form-action", []string{selfSrc}},
+				{"img-src", []string{selfSrc, "data:", "https:"}},
```
marwen-abid, line 278: ❓ Does this mean we can load images from `self` **or** any `data:` **or** any `https:` link ? Because if that's the case, I believe it opens the door for XSS attacks.

marwen-abid (approved):
I left a question below. Please feel free to merge if it's not a problem.

internal/serve/middleware/middleware.go
marcelosalloum: Hm, you're right.

`data:` should be enough, but I'll hold off from it since the FE seems to be loading correctly regardless of the console error.

We can reopen it if needed.

PR #616, opened 2025-04-08
SDP-1595 set default unique tenant in single tenant mode

dpohr:
### What

Added automatic default tenant resolution for single tenant mode:
- GetDefault if the singleTenant mode is active checks for the existing tenants and returns single active tenant
- Extended manager with single tenant property to control the behavior of the GetDefault 

### Why

In case the the SDP instance is set to single tenant mode and there is only one tenant in the database, we default to that tenant during the resolution process. 

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


stellar-multitenant/pkg/serve/serve.go
```diff
@@ -133,6 +134,7 @@ func handleHTTP(opts *ServeOptions) *chi.Mux {
 	// Authenticated Routes
 	mux.Group(func(r chi.Router) {
 		r.Use(middleware.BasicAuthMiddleware(opts.AdminAccount, opts.AdminApiKey))
+		r.Use(multitenantmiddleware.TenantResolutionMiddleware(opts.tenantManager, opts.SingleTenantMode))
```
marwen-abid, line 137: The issue with adding this logic to the middleware of the admin (tenant) API is that it will only be called if the end-user calls the `/tenants` API endpoints, which happens rarely (Tenant provisioning, patching, fetching). 

We want to make this behaviour as part of Tenant Resolution. That resolution happens in two places primarily : 
* `ResolveTenantFromRequestMiddleware` in `middleware.go` 
* `sep_24_auth_middleware`. 

Both operations use `tenantManager.GetDefault(ctx)`. So I suggest avoiding setting the default explicitly behind the scenes and just returning the one tenant if only one exists. 

Tenant resolution doesn't happen on every request, it only happens on login (then tenant ID gets stored in the jwt token).

internal/anchorplatform/sep24_auth_middleware_test.go
```diff
@@ -761,3 +761,40 @@ func Test_getCurrentTenant(t *testing.T) {
 		assert.Equal(t, &expectedTenant, currentTnt)
 	})
 }
+
+func Test_getCurrentTenant_WithSingleTenant(t *testing.T) {
```
marwen-abid, line 765: Thank you for adding these tests!

marwen-abid (approved):
This looks fantastic! Thank you @dpohr

PR #617, opened 2025-04-09
SDP-1596 add new broker type and deprecate enable_scheduler config

dpohr:
### What

### Why

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


marwen-abid (changes requested):
The main code change here is missing. We need to change 
`EventBrokerType` to support the new value `SCHEDULER`.  

We should also maintain support for the value `NONE`,  print a warning that this will be deprecated in future releases and recommend users to set that variable to `SCHEDULER`. 

Currently if I use the new config introduced the server fails to start. 
```
FATA[2025-04-14T18:29:45.274-07:00] Error setting values of config options: couldn't parse event broker type in event-broker-type: invalid event broker type  pid=16052
exit status 1
```

cmd/serve.go
```diff
@@ -656,15 +656,31 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
...
+			if serveOpts.EnableScheduler {
+				log.Ctx(ctx).Warn("The 'enable-scheduler' flag is deprecated. Please use 'event-broker-type=SCHEDULER' instead.")
 			}
-			if eventBrokerOptions.EventBrokerType != events.NoneEventBrokerType && serveOpts.EnableScheduler {
-				log.Ctx(ctx).Fatalf("Both Event Brokers and Scheduler are enabled. Please enable only one.")
+
+			switch eventBrokerOptions.EventBrokerType {
+			case events.KafkaEventBrokerType:
+				if serveOpts.EnableScheduler {
+					log.Ctx(ctx).Fatalf("Both Event Broker Type=KAFKA and enable-scheduler=true are set. Please use only one approach.")
```
marwen-abid, line 666: nit: I like that you used the cmd line arg name in the previous message. I think we should keep it consistent here. 
i.e. `Both event-broker-type=KAFKA and..`

cmd/serve.go
```diff
@@ -656,15 +656,31 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
...
+				}
+			case events.NoneEventBrokerType:
+				if !serveOpts.EnableScheduler {
+					log.Ctx(ctx).Fatalf("No background processing method is enabled. Please set event-broker-type=SCHEDULER or event-broker-type=KAFKA.")
+				} else {
+					log.Ctx(ctx).Warn("Using event-broker-type=NONE with enable-scheduler=true is deprecated. Please use event-broker-type=SCHEDULER instead.")
```
marwen-abid, line 672: nit: I believe this is redundant because of the log on line 660, but no strong opinion here. We can leave it.

cmd/serve.go
```diff
@@ -680,19 +696,18 @@ func (c *ServeCommand) Command(serverService ServerServiceInterface, monitorServ
 				if kafkaErr != nil {
 					log.Fatalf("error setting up consumers: %v", kafkaErr)
 				}
-			} else {
-				log.Ctx(ctx).Warn("Event Broker Type is NONE. Using Noop producer for logging events")
+			case events.NoneEventBrokerType:
+				log.Ctx(ctx).Warn("Event Broker Type is NONE (deprecated). Using NoopProducer for logging events.")
 				serveOpts.EventProducer = events.NoopProducer{}
+			case events.SchedulerEventBrokerType:
```
marwen-abid, line 702: 🚨 Unfortunately, this won't work. 

The naming we chose is confusing as `enable_scheduler` doesn't really enable all of the scheduler.

The scheduler needs to always be started, but the type of jobs that get added to the scheduler depend on that `enable_scheduler` flag. 

See here : https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/cmd/serve.go#L95

Kafka was primarily used to scale inter-service communication between the SDP and the TSS. Other jobs like the job that checks that anchor platform auth is properly enforced or that cancel payments within the SDP didn't need to be ported to Kafka.

README.md
```diff
@@ -216,6 +216,14 @@ Currently, the SDP only supports Kafka as an Event Broker even though it has bee
 > [!NOTE]  
 > In order to avoid concurrency issues, the SDP only supports one Event Broker or Background Jobs at a time.
 
+### Configuration Options
```
marwen-abid, line 219: `### Configuration options` is at the same header level as its parent `### Event Brokers & Background jobs`

marwen-abid (approved):
LGTM!

cmd/serve.go
marwen-abid: Addressed in cacea2dd4d440e27f33e26783e5d15b0c229af4c

PR #674, opened 2025-05-09
SDP-1616: Fix WebAuthn client_data_json field extraction

philipliu:
### What

This fixes how the `challenge` and `type` fields are extracted from the client data JSON. Instead of asking the user to provide the offsets of these fields, the client data JSON is parsed by `serde_json_core`.

### Why

Asking the client to provide the offsets can result in the user unintentionally authorizing transactions if the challenge offset provided is outside of the `challenge` field.

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/contracts/smart-wallet/src/test.rs
+++ b/contracts/smart-wallet/src/test.rs
@@ -79,15 +79,9 @@ fn sign(env: Env, challenge_hash: &[u8; 32], signing_key: &mut SigningKey) -> We
     raw_signature_bytes[0..32].copy_from_slice(r_bytes.as_slice());
     raw_signature_bytes[32..64].copy_from_slice(s_bytes.as_slice());
 
-    // Correctly determine the indices of the type and challenge in the client data JSON
-    let type_index = client_data_json.find("webauthn.get").unwrap() as u32;
-    let challenge_index = client_data_json.find(challenge_str).unwrap() as u32;
-
     WebAuthnCredential {
         client_data_json: Bytes::from_slice(&env, client_data_json_bytes),
         authenticator_data,
-        type_index,
-        challenge_index,
         signature: BytesN::from_array(&env, &raw_signature_bytes),
     }
 }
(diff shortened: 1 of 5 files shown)


contracts/smart-wallet/src/test.rs
```diff
@@ -289,51 +323,3 @@ fn test_webauthn_tampered_signature() {
 
     assert!(result.is_err());
 }
-
-#[test]
-fn test_webauthn_invalid_challenge_length_in_client_data() {
```
leighmcculloch, line 294: Is this test case, invalid length, still covered by another test? It sends like a valid case to test for still.

contracts/smart-wallet/src/webauthn.rs
```diff
@@ -2,16 +2,45 @@ use soroban_sdk::{contracttype, crypto::Hash, panic_with_error, Bytes, BytesN, E
...
+///
+/// - `origin`: ~100-200 bytes (`"origin":"https://example.com",`)
+///
+/// - `crossOrigin`: ~20 bytes (`"crossOrigin":false,`).
+///
+/// Total length: ~283 bytes.
+///
+/// This is a conservative estimate, as the actual length may vary based on the specific values used.
+/// The maximum length is set to 1024 bytes to accommodate any additional fields or whitespace.
+pub(crate) const MAX_CLIENT_DATA_JSON_LEN: usize = 1024;
```
leighmcculloch, line 34: Seems reasonable to me. 

I'm wondering what we could do to take away the need for this. I think the explicit size assumption could be removed if the to_buffer function was accompanied by a to_reader function that provided access to reading the string in chunks, and if serde_json_core supported accepting a read-like interface. Unfortunately serde_json_core doesn't support anything other than a full slice or string.

leighmcculloch (approved):

contracts/smart-wallet/src/test.rs
philipliu: You're right; there's a test that covers invalid challenge content, but having a test for invalid length is more robust.

contracts/smart-wallet/src/webauthn.rs
```diff
@@ -53,16 +78,19 @@ pub fn verify(
...
 ) {
-    // 1. Verify the Webauthn type
-    const WEBAUTHN_TYPE_GET_LEN: u32 = 12;
-
-    let type_slice = credential
+    // Parse the client data JSON
+    let client_data_json = credential
         .client_data_json
-        .slice(credential.type_index..(credential.type_index + WEBAUTHN_TYPE_GET_LEN));
+        .to_buffer::<MAX_CLIENT_DATA_JSON_LEN>();
```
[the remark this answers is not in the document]
philipliu: Well, I took this snippet from your code, so I can't take the credit here, but the interface does seem to work nicely for the JSON deserialization use case 🙂

PR #680, opened 2025-05-14
SDP-1614: Add wallet creation type to TSS transactions

philipliu:
### What

The `/embedded-wallets/create` endpoint will submit a new type of transaction for wallet creation to TSS. This makes the necessary changes to the `Transaction` model to support building the TSS workers and scheduled jobs/event handlers on top of it.

### Why

TSS will be submitting wallet creation transactions, but it's built for payment transactions and needs to be generalized to support other transaction types. This will require making changes to the transaction type, the transaction worker, and the payment sync jobs and event handlers. 

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


internal/transactionsubmission/store/transactions.go
```diff
@@ -65,7 +71,25 @@ type Transaction struct {
...
+	Amount      float64         `db:"amount"`
+	Destination string          `db:"destination"`
+	Memo        string          `db:"memo"`
+	MemoType    schema.MemoType `db:"memo_type"`
+}
+
+type WalletCreation struct {
+	PublicKey string `db:"public_key"`
+	WasmHash  string `db:"wasm_hash"`
+}
```
philipliu, line 86: @marcelosalloum brought up whether or not we want to prefix the columns with the transaction type in the database. 

I'm leaning towards not doing it since it doesn't really help too much with code readability IMO since we are accessing these fields through the `Transaction` struct, and it doesn't do much to prevent misuse of these columns. Instead, we are baking a lot of the safety into the `validate` method, which is called before the methods are inserted into the database.

PR #689, opened 2025-05-22
[SDP-1621] Implement wallet creation invite

philipliu:
### What

### Why

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


internal/services/send_receiver_wallets_invite_service.go
```diff
@@ -207,6 +214,14 @@ func (s SendReceiverWalletInviteService) SendInvite(ctx context.Context, receive
 	})
 }
 
+func (wdl *WalletDeepLink) updateDeepLink() error {
+	wdl.DeepLink = wdl.TenantBaseURL
```
philipliu, line 218: I'm not sure how we want to support different frontends (i.e., wallets created by SDP, but the frontend is not served by SDP). What we need to do is not replace the deep link for these wallets but include a token in the deep link. 

One approach is to add another field to the `wallets` model, `dynamic_deep_link`, when set to true, it replaces the deep link with the tenant.

marcelosalloum: I'd appreciate you explaining better what is the current link design and what do you envision for C-accounts.

IIUC, we could store embedded wallets in the DB in a similar fashion as we do today... the _self-hosted_ wallet, we could store the deepLink with a known value like `SELF` or `<SELF>, which would result in this very substitution you're doing here.

internal/services/wallets/wallets_futurenet.go
```diff
@@ -21,4 +21,16 @@ var FuturenetWallets = []data.Wallet{
...
+		Assets:   assets.AllAssetsTestnet,
+		Embedded: true,
+	},
+	{
+		Name:           "SDP Embedded Wallet",
+		DeepLinkSchema: "https://stellar.org",
+		Homepage:       "https://stellar.org",
+		Assets:         assets.AllAssetsTestnet,
+		Embedded:       true,
+	},
```
marcelosalloum, line 35: Shouldn't these have different names?

internal/services/wallets/wallets_testnet.go
```diff
@@ -30,4 +30,11 @@ var TestnetWallets = []data.Wallet{
 		Assets:      assets.AllAssetsTestnet,
 		UserManaged: true,
 	},
+	{
+		Name:           "SDP Embedded Wallet",
+		DeepLinkSchema: "https://stellar.org",
+		Homepage:       "https://stellar.org",
+		Assets:         assets.AllAssetsTestnet,
+		Embedded:       true,
+	},
```
marcelosalloum, line 39: [Not for this PR] WDYT about shielding the embedded wallet behind a feature flag? This may help us merge the C-accounts branch earlier.

marcelosalloum (commented):
Looking good!

I'd like us to decide about the deep_link substitution before approving this PR.

internal/services/wallets/wallets_futurenet.go
philipliu: This is a bug, there should be only one entry.

marcelosalloum (approved):
LGTM!

My only suggestion is renaming `Dynamic` to `SelfHosted`.

PR #716, opened 2025-06-18
SDP-1614: Handle Horizon errors in transaction building

philipliu:
### What

This adds Horizon error handling to the `BuildInnerTransaction` call in the TSS transaction worker.

### Why

Wallet creation calls Horizon's `AccountDetail` after simulating the transaction, which can fail. It needs to be wrapped in a Horizon error wrapper so that the transaction worker can properly mark it as retryable. 

### Known limitations

RPC simulation errors are currently not retried.

### Checklist

- [ ] Ready for production


internal/transactionsubmission/transaction_worker.go
```diff
@@ -431,12 +432,16 @@ func (tw *TransactionWorker) buildAndSignTransaction(ctx context.Context, txJob
...
-		err = fmt.Errorf("getting account detail: %w", err)
 		return nil, utils.NewHorizonErrorWrapper(err)
 	}
 
 	innerTx, err := tw.txHandler.BuildInnerTransaction(ctx, txJob, horizonAccount.Sequence, distributionAccount.Address)
 	if err != nil {
+		var hErr *utils.HorizonErrorWrapper
+		if errors.As(err, &hErr) {
+			return nil, hErr
+		}
```
marcelosalloum, line 443: Why don't you just return a new horizon error wrapper?

```suggestion
		return nil, utils.NewHorizonErrorWrapper(err)
```

PR #739, opened 2025-07-01
SDP-1722: Implement RPC error handling

philipliu:
### What

This adds comprehensive error handling and retry logic for RPC simulation failures. It introduces structured error categorization for different types of RPC simulation and implements retry behaviour based on error type, similar to how Horizon errors are handled today.

### Why

Wallet creation transactions rely on RPC for simulation before it is submitted. This happens when the transaction is being built, but errors at this point are not handled by TSS since they are never classified as retryable. 

### Known limitations

RPC error classification is a best-effort attempt. I'm not sure what the error looks like when RPC throttles the client.

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/wallet_creation_handler.go
+++ b/internal/serve/httphandler/wallet_creation_handler.go
@@ -1,6 +1,7 @@
 package httphandler
 
 import (
+	"crypto/ecdh"
 	"encoding/hex"
 	"errors"
 	"net/http"
@@ -31,8 +32,10 @@ func (r CreateWalletRequest) Validate() *httperror.HTTPError {
 	validator.Check(len(strings.TrimSpace(r.Token)) > 0, "token", "token should not be empty")
 	validator.Check(len(strings.TrimSpace(r.PublicKey)) > 0, "public_key", "public_key should not be empty")
 	validator.Check(len(strings.TrimSpace(r.CredentialID)) > 0, "credential_id", "credential_id should not be empty")
-	if _, err := hex.DecodeString(r.PublicKey); err != nil {
-		validator.AddError("public_key", "public_key should be a valid hex string")
+	if pk, err := hex.DecodeString(r.PublicKey); err != nil {
+		validator.AddError("public_key", "public_key is not a valid hex string")
+	} else if !isValidP256PublicKey(pk) {
+		validator.AddError("public_key", "public_key is not a valid uncompressed P256 public key")
 	}
 
 	if validator.HasErrors() {
@@ -40,6 +43,14 @@ func (r CreateWalletRequest) Validate() *httperror.HTTPError {
 	}
 
 	return nil
+}
+
+func isValidP256PublicKey(pubKeyBytes []byte) bool {
+	if len(pubKeyBytes) != 65 || pubKeyBytes[0] != 0x04 {
+		return false
+	}
+	_, err := ecdh.P256().NewPublicKey(pubKeyBytes)
+	return err == nil
 }
 
 type WalletResponse struct {
--- a/internal/serve/httphandler/wallet_creation_handler_test.go
+++ b/internal/serve/httphandler/wallet_creation_handler_test.go
@@ -28,12 +28,12 @@ func Test_WalletCreationHandler_CreateWallet(t *testing.T) {
 	rr := httptest.NewRecorder()
 	requestBody, _ := json.Marshal(CreateWalletRequest{
 		Token:        "123",
-		PublicKey:    "04f5",
+		PublicKey:    "04f5549c5ef833ab0ade80d9c1f3fb34fb93092503a8ce105773d676288653df384a024a92cc73cb8089c45ed76ed073433b6a72c64a6ed23630b77327beb65f23",
 		CredentialID: "test-credential-id",
 	})
 	ctx := context.Background()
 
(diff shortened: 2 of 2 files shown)


internal/transactionsubmission/transaction_worker.go
```diff
@@ -162,66 +163,57 @@ func (tw *TransactionWorker) runJob(ctx context.Context, txJob *TxJob) error {
...
-//   - 429: Too Many Requests
-//   - 400 - tx_insufficient_fee: Bad Request
+// Errors that trigger the pause/jitter mechanism at TransactionProcessingLimiter:
+//
+//	Horizon: 504 (Timeout), 429 (Too Many Requests), 400 tx_insufficient_fee
+//	RPC: Network errors, Resource errors
 //
 // Errors marked as definitive error, that won't be resolved with retries:
-//   - 400: with any of the transaction error codes [tx_bad_auth, tx_bad_auth_extra, tx_insufficient_balance]
-//   - 400: with any of the operation error codes [op_bad_auth, op_underfunded, op_src_not_authorized, op_no_destination, op_no_trust, op_line_full, op_not_authorized, op_no_issuer]
```
marcelosalloum, line 176: Was this comment incorrect? Should we continue mentioning the errors `[op_bad_auth, op_underfunded, op_src_not_authorized, op_no_destination, op_no_trust, op_line_full, op_not_authorized, op_no_issuer]`?

internal/transactionsubmission/transaction_worker.go
```diff
@@ -467,22 +484,23 @@ func (tw *TransactionWorker) buildAndSignTransaction(ctx context.Context, txJob
...
 
-	return feeBumpTx, nil
+	return feeBumpTx, false, nil
 }
 
 func (tw *TransactionWorker) submit(ctx context.Context, txJob *TxJob, feeBumpTx *txnbuild.FeeBumpTransaction) error {
 	resp, err := tw.engine.HorizonClient.SubmitFeeBumpTransactionWithOptions(feeBumpTx, horizonclient.SubmitTxOpts{SkipMemoRequiredCheck: true})
 	if err != nil {
-		err = tw.handleFailedTransaction(ctx, txJob, resp, utils.NewHorizonErrorWrapper(err))
+		txErr := &utils.HorizonTransactionError{HorizonErrorWrapper: utils.NewHorizonErrorWrapper(err)}
```
marcelosalloum, line 502: Why do we need two levels of wrappers here?

internal/transactionsubmission/transaction_worker.go
```diff
@@ -353,8 +370,11 @@ func (tw *TransactionWorker) processTransactionSubmission(ctx context.Context, t
 	}
 
 	// STEP 2: prepare transaction for processing
-	feeBumpTx, err := tw.prepareForSubmission(ctx, txJob)
+	feeBumpTx, handled, err := tw.prepareForSubmission(ctx, txJob)
 	if err != nil {
+		if handled {
+			return nil
+		}
```
marcelosalloum, line 377: 💡  Rather than passing `handled` all the way up just to ignore the error, it'd be better to just return a nil error.

So basically when `handlePreparationError` handles the error, you don't pass it up and just return nil. WDYT?

philipliu: Nice, didn't think of this 🙂

internal/stellar/rpc.go
```diff
@@ -16,9 +20,143 @@ type RPCOptions struct {
...
+func (e *SimulationError) Unwrap() error {
+	return e.Err
+}
+
+func (e *SimulationError) IsRetryable() bool {
+	switch e.Type {
+	case SimulationErrorTypeNetwork, SimulationErrorTypeResource:
+		return true
+	case SimulationErrorTypeTransactionInvalid, SimulationErrorTypeAuth, SimulationErrorTypeContractExecution, SimulationErrorTypeUnknown:
+		return false
```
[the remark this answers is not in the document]
philipliu: We should remove the default case instead so that if we add new error types in the future, the compiler would force us to handle them here.

internal/transactionsubmission/transaction_worker.go
philipliu: No, it should be included. I missed this when I rewrote the comment to include the RPC errors.

internal/transactionsubmission/transaction_worker.go
philipliu: Yeah, it's really just delegating to the inner wrapper. I'll remove it.

internal/transactionsubmission/transaction_worker.go
marcelosalloum: SGTM!

internal/transactionsubmission/engine/tx_processing_limiter.go
```diff
@@ -98,15 +88,15 @@ func (tpl *TransactionProcessingLimiterImpl) adjustLimitIfNeeded() {
...
 			}
 		}
-	case *utils.RPCTransactionError:
+	case *utils.RPCErrorWrapper:
 		if e.IsRPCError() {
-			if shouldAdjustLimitForRPCError(e.RPCErrorWrapper) {
+			if shouldAdjustLimitForRPCError(e) {
 				tpl.adjustLimitIfNeeded()
 			}
 		}
```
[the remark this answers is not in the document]
philipliu: I keep forgetting to use `errors.As` 😅

PR #754, opened 2025-07-04
SDP-1738: Associate wallet creation token with receiver

philipliu:
### What

This updates the wallet creation tokens with the receiver for which it was generated.

### Why

Visibility into whether a receiver has created a wallet using a token.

### Known limitations

N/A

### Checklist

- [ ] Ready for production


internal/data/embedded_wallet.go
```diff
@@ -83,6 +84,7 @@ func EmbeddedWalletColumnNames(tableReference, resultAlias string) string {
 			"wallet_status",
 			"contact_type",
 			"receiver_contact",
+			"receiver_id",
```
marwen-abid, line 87: Can ReceiverID be null ?

internal/data/embedded_wallet.go
```diff
@@ -136,11 +138,57 @@ func (ew *EmbeddedWalletModel) GetByCredentialID(ctx context.Context, sqlExec db
 	return &wallet, nil
 }
 
+// GetByReceiverID returns all embedded wallets for a specific receiver
```
marwen-abid, line 141: nit: Why not just have one method instead of two, and use a variadic operator. 
```
func (ew *EmbeddedWalletModel) GetByReceiverIDs(
     ctx context.Context, 
     sqlExec db.SQLExecuter, 
     receiverIDs ...string) ([]EmbeddedWallet, error) {

	if len(receiverIDs) == 0 {
		return []EmbeddedWallet{}, nil
	}
```

marwen-abid: If you want to keep them separate you can also do the following ? 
```
func (ew *EmbeddedWalletModel) GetByReceiverID(ctx context.Context, sqlExec db.SQLExecuter, receiverID string) ([]EmbeddedWallet, error) {
	return ew.queryByReceiverIDs(ctx, sqlExec, []string{receiverID})
}
```

internal/serve/httphandler/receiver_handler.go
```diff
@@ -215,7 +231,16 @@ func (rh ReceiverHandler) GetReceivers(w http.ResponseWriter, r *http.Request) {
 			return nil, fmt.Errorf("error retrieving receiver wallets: %w", err)
 		}
 
-		receiversResponse := rh.buildReceiversResponse(receivers, receiversWallets)
+		var allEmbeddedWallets []data.EmbeddedWallet
+		for _, receiverID := range receiverIDs {
+			embeddedWallets, walletsErr := rh.Models.EmbeddedWallets.GetByReceiverID(ctx, dbTx, receiverID)
```
marwen-abid, line 236: This will result in an n+1 query if going to the database inside a loop. This doesn't matter if this code is only intended for the demo, but if it's going to eventually make it to prod I would be in favor of changing it by making one query to fetch all the embedded wallets.

marwen-abid (approved):
LGTM

internal/data/embedded_wallet.go
philipliu: Oops, thanks for catching this!

PR #782, opened 2025-07-30
Fix: Embedded wallet API auth

philipliu:
### What

This fixes the API key authentication for the embedded wallet endpoints. It was broken since it created a new `mux.Group`, so the API auth middleware was not properly applied. It also narrows the permissions for each endpoint.

### Why

Cannot access these endpoints even with a valid API key

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/serve.go
+++ b/internal/serve/serve.go
@@ -572,17 +572,26 @@ func handleHTTP(o ServeOptions) *chi.Mux {
 
 		// Embedded wallet routes (only if feature is enabled)
 		if o.EnableEmbeddedWallets && o.EmbeddedWalletService != nil {
-			mux.Group(func(r chi.Router) {
-				walletCreationHandler := httphandler.WalletCreationHandler{
-					EmbeddedWalletService: o.EmbeddedWalletService,
-				}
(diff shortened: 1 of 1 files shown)


marwen-abid (approved):

internal/serve/serve.go
```diff
@@ -572,17 +572,26 @@ func handleHTTP(o ServeOptions) *chi.Mux {
...
-					r.Post("/", walletCreationHandler.CreateWallet)
+				)).Group(func(r chi.Router) {
 					r.Get("/{credentialID}", walletCreationHandler.GetWallet)
 					r.Get("/status/{token}", walletCreationHandler.GetWalletStatus)
+				})
+
+				// Write operations
+				r.With(middleware.RequirePermission(
+					data.WriteDisbursements,
+					middleware.AnyRoleMiddleware(authManager, data.GetAllRoles()...),
```
marwen-abid, line 592: Should all roles be able to create a wallet ?

philipliu: It should only be accessible via the API for the demo, but since it's also for a demo, I think it's fine to leave it as is.

PR #784, opened 2025-08-01
[DRAFT] address multisig for the account responsible for recovering the embedded accounts

marcelosalloum:
### What

### Why

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


internal/transactionsubmission/wallet_creation_transaction_handler.go
```diff
@@ -194,17 +192,6 @@ func (h *WalletCreationTransactionHandler) BuildInnerTransaction(ctx context.Con
 		return nil, applyErr
 	}
 
-	txParams.BaseFee = h.calculateAdjustedBaseFee(simulationResponse)
-
```
marcelosalloum, line 198: The fee adjustment was a workaround needed due to a design restriction on the wallet-backend repo. I removed the call because it's not needed for us here.

internal/transactionsubmission/wallet_creation_transaction_handler.go
```diff
@@ -194,17 +192,6 @@ func (h *WalletCreationTransactionHandler) BuildInnerTransaction(ctx context.Con
...
-	channelAccount, err := h.engine.HorizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: txJob.ChannelAccount.PublicKey})
-	if err != nil {
-		return nil, utils.NewHorizonErrorWrapper(err)
-	}
-	txParams.SourceAccount = &txnbuild.SimpleAccount{
-		AccountID: txJob.ChannelAccount.PublicKey,
-		Sequence:  channelAccount.Sequence,
-	}
-
```
marcelosalloum, line 207: Removed this part because we already had the sequence setup a few lines above, so this is not needed.

helmchart/sdp/values.yaml
```diff
@@ -199,8 +199,10 @@ sdp:
...
-  ## @extra sdp.configMap.data.EMBEDDED_WALLETS_WASM_HASH The WASM hash of the smart contract for embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true".
-  ## @extra sdp.configMap.data.EMBEDDED_WALLETS_RECOVERY_ADDRESS The recovery address for embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true".
+## @extra sdp.configMap.data.EMBEDDED_WALLETS_WASM_HASH The WASM hash of the smart contract for embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true".
+## @extra sdp.configMap.data.EMBEDDED_WALLETS_RECOVERY_ADDRESS The recovery address for embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true".
+## @extra sdp.configMap.data.EMBEDDED_WALLETS_MULTISIG_ADMIN_ADDRESS The admin address for multisig configuration of embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true" and multisig is enabled.
+## @extra sdp.configMap.data.EMBEDDED_WALLETS_MULTISIG_SECONDARY_SIGNER_ADDRESS The secondary signer address for multisig configuration of embedded wallets. Required when ENABLE_EMBEDDED_WALLETS is "true" and multisig is enabled.
```
philipliu, line 205: Are these used?

internal/serve/httphandler/cosign_recovery_handler.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if invokeContractOp.FunctionName != RotateSignerFnName {
+		return nil, fmt.Errorf("wrong function being called: %s != %s", invokeContractOp.FunctionName, RotateSignerFnName)
+	}
+
+	for _, sorobanAuth := range invokeHostOp.Auth {
+		if len(sorobanAuth.RootInvocation.SubInvocations) > 0 {
+			return nil, errors.New("contract operation has subinvocations which are not allowed")
+		}
+	}
```
philipliu, line 132: Should we also validate the contents of the auth entries? Such as:
- Expecting an entry for the recovery address
- The arguments matching the invoke operation

internal/serve/httphandler/cosign_recovery_handler.go
```diff
@@ -0,0 +1,154 @@
...
+		httperror.InternalError(ctx, err.Error(), err, nil).Render(rw)
+		return
+	}
+
+	resp := TransactionPayload{TransactionXDR: txXDR}
+	httpjson.Render(rw, resp, httpjson.JSON)
+}
+
+// parseAndValidateTransaction parses the transaction XDR and performs a series of security validations:
+// - The transaction must have exactly one operation
```
philipliu, line 78: Should we also validate the transaction source account?

internal/transactionsubmission/services/recovery_multisig_setup_service.go
```diff
@@ -0,0 +1,164 @@
...
+		SourceAccount: &recoveryAccount,
+		Operations: []txnbuild.Operation{
+			&txnbuild.SetOptions{
+				// Set thresholds to 2
+				LowThreshold:    txnbuild.NewThreshold(2),
+				MediumThreshold: txnbuild.NewThreshold(2),
+				HighThreshold:   txnbuild.NewThreshold(2),
+				// Add signer with weight=1
+				Signer: &txnbuild.Signer{
+					Address: opts.CosignerPublicKey,
```
philipliu, line 81: If we somehow pass the wrong cosigner public key, we wouldn't be able to use this command to fix it since we only sign with the master key. The risk of this is low, and we can always fix it by manually crafting and submitting the transaction somewhere else, but I wanted to call this out.

internal/transactionsubmission/services/recovery_multisig_setup_service.go
```diff
@@ -0,0 +1,164 @@
...
+		return fmt.Errorf("transaction failed with ResultXdr=%s", resp.ResultXdr)
+	}
+
+	log.Ctx(ctx).Infof("✅ Account %s now requires 2 signatures for all operations", recoveryMasterKP.Address())
+	return nil
+}
+
+// isAccountMultisigConfigured checks if the account has a multisig configured with the given cosigner public key.
+// It asserts that the account has thresholds [2,2,2] and signers weights [cosigner=1,admin=1].
+func isAccountMultisigConfigured(_ context.Context, account horizon.Account, cosignerPublicKey string) bool {
```
philipliu, line 128: Should we also call this function after submitting the transaction to check that the account is set up correctly?

internal/transactionsubmission/services/recovery_multisig_setup_service.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/stellar/go/protocols/horizon"
+	"github.com/stellar/go/strkey"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/txnbuild"
+
+	"github.com/stellar/stellar-disbursement-platform-backend/internal/transactionsubmission/utils"
+)
+
+type RecoveryMultisigSetupOptions struct {
+	MasterPublicKey   string
```
philipliu, line 18: Is `MasterPublicKey` used?

internal/transactionsubmission/services/recovery_multisig_setup_service.go
marcelosalloum: That's true. We'll add it to the handbook and test things in advance to make sure we don't incur on these errors.

internal/transactionsubmission/services/recovery_multisig_setup_service.go
marcelosalloum: I was defaulting to the address gathered from the seed, but I made the change to use this one instead: https://github.com/stellar/stellar-disbursement-platform-backend/pull/784/commits/16255177655d3109032fde25f28a6ab97ea575dd

internal/transactionsubmission/services/recovery_multisig_setup_service.go
marcelosalloum: Good question... TBH I don't think it's necessary. We don't do this for any other kind of transaction, instead we trust we've built the transactions correctly on our end and we ensure that through 2 means:
1. using unit tests to assert the content of the transaction. Added in https://github.com/stellar/stellar-disbursement-platform-backend/pull/784/commits/d8eaea55876f63800aa69d3561b27f662aadbaa6
2. ensuring the transaction was successful.

LMK if you still think we should do the extra check thoug, I'm open to discussing it further.

internal/serve/httphandler/cosign_recovery_handler.go
marcelosalloum: The client has the freedom to set the SDP's recovery signer as either the tx.SourceAccount or the op.SourceAccount.

If not properly set, the worst that could happen is a failed transaction, right? This method is more focused on validating security issues though.

Would you be ok moving forward without that check? LMK otherwise.

helmchart/sdp/values.yaml
marcelosalloum: Good catch! The variables were wrong. I updated the code by leaving just `EMBEDDED_WALLETS_RECOVERY_MASTER_PRIVATE_KEY`.

We also have `EMBEDDED_WALLETS_RECOVERY_COSIGNER_PUBLIC_KEY` but that one is only used when running the CLI command, so I won't add it to the deployment values.

internal/serve/httphandler/cosign_recovery_handler.go
philipliu: Yeah, it's overly defensive since the API is authenticated and access is limited to the demo backend only. I'm good with keeping it as is.

internal/transactionsubmission/services/recovery_multisig_setup_service.go
philipliu: I think it's fine to leave as is. Whether we include this additional check or not, we should still verify the account was set up correctly using something like stellar.expert, so in the end it's redundant.

PR #791, opened 2025-08-06
SDP 1793 add asset filter by trustline

dpohr:
### What

Added trustline-based filtering to the GET /assets endpoint to prevent Direct Payment failures due to missing trustlines. The changes include:

New Features:
- Query Parameter: GET /assets?hasTrustline={true|false} - Filters assets based on distribution account trustline status
- Response Enhancement: When hasTrustline parameter is provided, response includes has_trustline boolean field for each asset
- Backward Compatibility: Default behavior (no hasTrustline parameter) returns all assets without trustline information

### Why

Direct Payments using assets without trustlines were failing at runtime, but these assets were still displayed in the UI, leading to poor user experience.

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/assets_handler.go
+++ b/internal/serve/httphandler/assets_handler.go
@@ -112,9 +112,6 @@ func (c AssetsHandler) getTrustlineInfo(
 	account *schema.TransactionAccount,
 	asset data.Asset,
 ) (bool, *float64, error) {
-	if asset.IsNative() {
-		return true, nil, nil
-	}
 
 	if !account.IsStellar() {
 		for _, networkAssets := range circle.AllowedAssetsMap {
--- a/internal/serve/httphandler/assets_handler.go
+++ b/internal/serve/httphandler/assets_handler.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strconv"
 	"strings"
 
 	"github.com/go-chi/chi/v5"
@@ -19,13 +18,13 @@ import (
 	"github.com/stellar/go/txnbuild"
 
 	"github.com/stellar/stellar-disbursement-platform-backend/db"
-	"github.com/stellar/stellar-disbursement-platform-backend/internal/circle"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/data"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/serve/httperror"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/serve/validators"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/services"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/transactionsubmission/engine"
 	tssUtils "github.com/stellar/stellar-disbursement-platform-backend/internal/transactionsubmission/utils"
+	"github.com/stellar/stellar-disbursement-platform-backend/internal/utils"
 	"github.com/stellar/stellar-disbursement-platform-backend/pkg/schema"
 )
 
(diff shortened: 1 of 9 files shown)


internal/serve/httphandler/assets_handler.go
```diff
@@ -57,14 +66,94 @@ func (c AssetsHandler) GetAssets(w http.ResponseWriter, r *http.Request) {
...
+
+		httpjson.Render(w, responseAssets, httpjson.JSON)
+		return
+	}
+
 	httpjson.Render(w, assets, httpjson.JSON)
 }
 
+// checkTrustlineExists checks if the distribution account has a trustline for the given asset.
+func (c AssetsHandler) checkTrustlineExists(
```
marwen-abid, line 108: You can use `s.distributionAccountService.GetBalance(ctx, &distributionAccount, usdcAsset)` and `s.distributionAccountService.GetBalances(...)` 

I did it for my PR here that checks for a similar thing https://github.com/stellar/stellar-disbursement-platform-backend/blob/2bacf252c1c9d9cafa8dfb67b8afbf75015af06f/internal/bridge/service.go#L369

marwen-abid: @dpohr Let's also return the Balance when it's available as we will need it for a direct payment change that's coming later.

internal/serve/httphandler/assets_handler.go
```diff
@@ -57,14 +68,75 @@ func (c AssetsHandler) GetAssets(w http.ResponseWriter, r *http.Request) {
...
+	}
+
 	httpjson.Render(w, assets, httpjson.JSON)
 }
 
+func (c AssetsHandler) getTrustlineInfo(
+	ctx context.Context,
+	account *schema.TransactionAccount,
+	asset data.Asset,
+) (bool, *float64, error) {
```
marwen-abid, line 114: There are a couple of issues here : 
1. The error return parameter is always nil 
2. `c.DistributionAccountService.GetBalance(ctx, account, asset)` works for both Stellar and Circle, so you don't need special handling for it.  This would allow us to also show the balances for Circle. 

I think we should do a couple of things: 
1. add a special error to GetBalance to return `ErrNoBalanceForAsset` when there is actually no balance (specifically for the Stellar implementation). 
2. Refactor this to something simpler 
```go
// getBalanceInfo retrieves the trustline information for a given asset and account.
func (c AssetsHandler) getBalanceInfo(
	ctx context.Context,
	account *schema.TransactionAccount,
	asset data.Asset,
) (bool, *float64, error) {
	balance, err := c.DistributionAccountService.GetBalance(ctx, account, asset)
	if err != nil {
		if errors.Is(err, services.ErrNoBalanceForAsset) {
			return false, nil, nil
		} else {
			return false, nil, fmt.Errorf("getting balance for asset %s-%s: %w", asset.Code, asset.Issuer, err)
		}
	}
	return true, &balance, nil
}
```

internal/serve/httphandler/assets_handler.go
```diff
@@ -33,18 +36,26 @@ var errCouldNotRemoveTrustline = errors.New("could not remove trustline")
...
+	data.Asset
+	HasTrustline bool     `json:"has_trustline"`
+	Balance      *float64 `json:"balance,omitempty"`
+}
+
 // GetAssets returns a list of assets.
 func (c AssetsHandler) GetAssets(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	walletID := strings.TrimSpace(r.URL.Query().Get("wallet"))
+	hasTrustlineParam := strings.TrimSpace(r.URL.Query().Get("hasTrustline"))
```
marwen-abid, line 58: We should use snake case for query parameters to be consistent with other API endpoints. 
If you see occurrences of camel case please raise it so that we can address it.

marwen-abid (approved):
LGTM! Feel free to merge after addressing the issues here.

PR #807, opened 2025-08-14
SDP-1797 update create receiver validation

dpohr:
### What

updated validations to return more detailed message on FE

### Why

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/receiver_handler.go
+++ b/internal/serve/httphandler/receiver_handler.go
@@ -5,11 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/go-chi/chi/v5"
 
-	"github.com/stellar/go/strkey"
 	"github.com/stellar/go/support/render/httpjson"
 
 	"github.com/stellar/stellar-disbursement-platform-backend/db"
@@ -17,7 +15,6 @@ import (
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/serve/httperror"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/serve/httpresponse"
 	"github.com/stellar/stellar-disbursement-platform-backend/internal/serve/validators"
-	"github.com/stellar/stellar-disbursement-platform-backend/internal/utils"
 	"github.com/stellar/stellar-disbursement-platform-backend/pkg/schema"
 )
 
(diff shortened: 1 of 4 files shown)


internal/serve/httphandler/receiver_handler.go
```diff
@@ -246,11 +171,39 @@ func (rh ReceiverHandler) CreateReceiver(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if err = req.Validate(); err != nil {
-		httperror.BadRequest("validation error", err, nil).Render(w)
+	validatorReq := &validators.CreateReceiverRequest{
```
marwen-abid, line 174: I think it's better not to duplicate data structures just to avoid import cycles. There is obviously a design issue here that we should address and the duplication is a short term solution. 

I think one way is we introduce a third package (e.g. `api` or a better name) where we keep the "DTO" type structs that are used by other packages. You can do it just for `receiver` for now and I'll move the rest in another ticket. 

```
 Example structure:
  internal/serve/
  ├── api/
  │   ├── receiver.go     # CreateReceiverRequest, GetReceiverResponse
  │   ├── payment.go      
  │   └── common.go       # Shared 
  ├── httphandler/
  └── validators/
```

internal/serve/validators/receiver_validator_test.go
```diff
@@ -0,0 +1,230 @@
...
+			{
+				Address: "  GCQFMQ7U33ICSLAVGBJNX6P66M5GGOTQWCRZ5Y3YXYK3EB3DNCWOAD5K  ",
+				Memo:    "  my-precious  ",
+			},
+		},
+	}
+
+	validator.ValidateCreateReceiverRequest(request)
+
+	assert.Equal(t, "merry@buckland.me", request.Email)
```
marwen-abid, line 224: nit: better avoid real domains even in tests. We never know what will be hosted on `buckland.me` :)

internal/serve/validators/receiver_validator_test.go
```diff
@@ -0,0 +1,230 @@
...
+					},
+				},
+			},
+			expectError: true,
+			errorFields: []string{"verifications[0].type"},
+		},
+		{
+			name: "invalid date of birth format",
+			request: CreateReceiverRequest{
+				Email:      "samwise@shire.me",
```
marwen-abid, line 169: Same for `shire.me`. This one actually redirects to a print shop

internal/serve/httphandler/receiver_handler_test.go
```diff
@@ -1725,14 +1727,150 @@ func Test_ReceiverHandler_CreateReceiver_Validation(t *testing.T) {
...
+			req := httptest.NewRequest("POST", "/receivers", bytes.NewBuffer(reqBody))
+			req.Header.Set("Content-Type", "application/json")
+
+			w := httptest.NewRecorder()
+
+			r.ServeHTTP(w, req)
+
+			assert.Equal(t, tc.expectedStatus, w.Code)
+
+			var errorResponse struct {
```
marwen-abid, line 1855: You can use 

```go
var errorResponse httperror.HTTPError
err = json.Unmarshal(w.Body.Bytes(), &errorResponse)
require.NoError(t, err)

assert.Equal(t, "validation error", errorResponse.Error())
``

PR #826, opened 2025-08-22
Extend storage TTL on `__check_auth`

philipliu:
### What

This extends the instance storage by a month whenever the TTL falls below a 1-week threshold.

### Why

Prevent the contract code and instance from being archived.

### Known limitations

N/A

### Checklist

- [x] Ready for production


marcelosalloum (approved):

leighmcculloch:
> @leighmcculloch: do you think this is a good practice, in general?

@marcelosalloum Good question. It isn't obvious to me that a contract should do this because it is likely to be ineffective due to being based around specific assumptions that may or may not hold for a user.

> This extends the instance storage by a month whenever the TTL falls below a 1-week threshold.

If the goal is to prevent archiving, then the change in this pull request will only succeed at preventing archiving if the user uses their account to authorise actions in the last 7 days of every month. That seems awfully specific. I'd be surprised if we can predict a users activity so well as to assume they'll use it in the last week of every month.

> Prevent the contract code and instance from being archived.

What problem is the code and instance being archived causing?

PR #848, opened 2025-09-03
SDP-1837: Add endpoint for patch receiver wallet

JiahuiWho:
### What

This PR add endpoint `PATCH /{receiver_id}/wallets/{receiver_wallet_id}` to handle user managed wallet editing
1. `PatchReceiverWallet` for validate request, editing, error handling. Test case added
2.  ~~Moved `parseHttpConflictErrorIfNeeded` to `httperror.go` to reuse the duplicate wallet address error handling~~
3. Moved PostgreSQL constraint violation error handling from HTTP layer to data layer

### Why

Backend work for allowing editing to receiver known stellar address.

### Known limitations

N/A

### Checklist

- [ ] Ready for production


internal/serve/httphandler/receiver_wallets_handler.go
```diff
@@ -160,3 +162,107 @@ func (h ReceiverWalletsHandler) validateAndUpdateStatus(ctx context.Context, rec
...
+		if !currentReceiverWallet.Wallet.UserManaged {
+			return nil, httperror.BadRequest("Cannot edit stellar address for non-user-managed wallet", nil, nil)
+		}
+
+		// 2: Prepare the wallet update with new address and optional memo
+		walletUpdate := data.ReceiverWalletUpdate{
+			StellarAddress: patchRequest.StellarAddress,
+		}
+
+		if strings.TrimSpace(patchRequest.StellarMemo) != "" {
```
marwen-abid, line 227: We should validate the Memo and detect it's type before this step. We support all memo types normally so we shouldn't hardcode the type to ID. 
I know this is how the Create Receiver is implemented but there is a bug for that now. https://stellarorg.atlassian.net/browse/SDP-1845

internal/serve/httperror/httperror.go
```diff
@@ -133,3 +136,41 @@ func UnprocessableEntity(msg string, originalErr error, extras map[string]interf
...
+	}
+
+	// Handle unique constraint violations (email, phone number)
+	if pqErr.Code == pgUniqueViolation {
+		allowedConstraints := []string{"receiver_unique_email", "receiver_unique_phone_number"}
+		if !slices.Contains(allowedConstraints, pqErr.Constraint) {
+			return nil
+		}
+
+		fieldName := strings.Replace(pqErr.Constraint, "receiver_unique_", "", 1)
```
marwen-abid, line 159: I like that you extracted this error handling logic to be able to reuse it. 

The issue is we don't really want to have the handler layer know too much about the pg errors. That's a leaky abstraction. 

We should extract the logic that detects the conflict and stick it in the data package. Basically we should not be parsing `pq.Error` type of errors in handler, instead we should parse those in data package then return a custom error.

PR #861, opened 2025-09-10
SDP-1851 add mfa/captcha to the organization settings

dpohr:
### What

Add MFA and CAPTCHA enable settings to the organization level.

UI changes: https://github.com/stellar/stellar-disbursement-platform-frontend/pull/345

### Why

Currently, MFA and RECAPTCHA in the SDP are controlled by environment host-level flags that apply to all tenants, added organization level settings with a fallback to the environment settings

### Known limitations

[TODO or N/A]

### Checklist

- [ ] Ready for production


stellar-multitenant/internal/provisioning/manager.go
```diff
@@ -39,6 +39,8 @@ type ProvisionTenant struct {
 	BaseURL                 string
 	NetworkType             string
 	DistributionAccountType schema.AccountType
+	MFAEnabled              *bool
+	CAPTCHAEnabled          *bool
```
marwen-abid, line 43: It looks like we never pass a nil value here. If that's the case, this doesn't need to be a pointer.

internal/serve/httphandler/login_handler.go
```diff
@@ -40,15 +40,56 @@ type LoginHandler struct {
 	MFADisabled        bool
 }
 
-func (h LoginHandler) validateRequest(req LoginRequest, headers http.Header) *httperror.HTTPError {
+// isMFADisabled checks if MFA is disabled considering both org and environment settings
+func (h LoginHandler) isMFADisabled(ctx context.Context) bool {
+	// If environment disables MFA globally, it's disabled
```
marwen-abid, line 45: Isn't this supposed to be the other way around ? The Organization (more granular) level takes precedence, and if that's NIL, we default to the global one ?

internal/serve/httphandler/login_handler.go
```diff
@@ -40,15 +40,56 @@ type LoginHandler struct {
...
+		return !*org.MFAEnabled
+	}
+
+	return h.MFADisabled
+}
+
+// isCAPTCHADisabled checks if CAPTCHA is disabled considering both org and environment settings
+func (h LoginHandler) isCAPTCHADisabled(ctx context.Context) bool {
+	// If environment disables CAPTCHA globally, it's disabled
+	if h.ReCAPTCHADisabled {
```
marwen-abid, line 65: Same here, my understanding is that org level takes precedence.

internal/serve/httphandler/login_handler.go
```diff
@@ -40,15 +40,56 @@ type LoginHandler struct {
...
+		return true
+	}
+
+	org, err := h.Models.Organizations.Get(ctx)
+	if err != nil {
+		return h.ReCAPTCHADisabled
+	}
+
+	if org.CAPTCHAEnabled != nil {
+		return !*org.CAPTCHAEnabled
```
marwen-abid, line 75: 💡 From a cognitive pov it seems to add complexity to have to reason both on Disabled and Enabled recaptcha at the same time. Maybe it makes sense to switch the org level capatcha to be `CAPTCHADisabled` to be consistent. (same for MFA).

internal/serve/httphandler/login_handler.go
dpohr: Okay, the initial idea was that if the mfa was disabled on deployment, it means that its turned off completely, but if its enabled the orgs can decide

stellar-multitenant/internal/provisioning/manager.go
dpohr: this is used for the update query, we need to distinguish, do we have value for update or not, I faced an issue previously when non pointer false was treated as "nothing to change" and was ignored by the DB

marwen-abid (approved):

PR #887, opened 2025-10-02
SDP-1862: Add public key to embedded wallets table

philipliu:
### What

Persist the passkey's public key in the embedded wallets table so that it can be used to verify the WebAuthn challenge signed by the frontend.

### Why

Wallet authentication

### Known limitations

N/A

### Checklist

- [ ] Ready for production


philipliu:
> Quick question : will the PR preview instance pick up the new embedded_wallets.public_key column? Goose tends to skip a migration if it sees the same version already applied (with previous checksum)

I haven't confirmed but I assume so since the preview uses an ephemeral database. The reason I'm editing an existing migration file is that this is being merged into a feature branch, otherwise I would be creating a new migration file.

PR #890, opened 2025-10-06
SDP-1862: Implement passkey registration + authentication

philipliu:
### What

This implements the passkey registration + authentication flows. The registration flow is called to get the credential ID, and public key needed to call the embedded wallet API. The authentication flow is called to get embedded wallet associated with a passkey and to generate the JWT the wallet will use to access authenticated APIs.

The HTTP handler + dependency injection will be added in a follow-up PR.

### Why

Implements the passkey related APIs for the wallet frontend.

### Known limitations

N/A

### Checklist

- [ ] Ready for production


internal/wallet/webauthn_service.go
```diff
@@ -0,0 +1,346 @@
...
+			publicKey:    embeddedWallet.PublicKey,
+			wallet:       embeddedWallet,
+		}, nil
+	}
+
+	// FinishPasskeyLogin verifies the authentication response from the client:
+	// 1. Calls getUserForCredential handler to look up the user and their public key
+	// 2. Validates the cryptographic signature using the stored public key
+	// 3. Returns the authenticated user if signature verification succeeds
+	user, _, err := w.webAuthn.FinishPasskeyLogin(getUserForCredential, session, request)
```
JiahuiWho, line 300: The parse call at line 267 consumes `request.Body` right? And the here `body` will be empty and the library will fail to deserialize the assertion?

philipliu: Good catch. This should be using `ValidatePasskeyLogin` instead.

PR #903, opened 2025-10-13
SDP-1880: Block enabling embedded wallet when feature flag disabled

JiahuiWho:
### What

  - Add an embedded data-layer filter so `/wallets` queries can drop the embedded provider when the feature flag is off.
  - Pass the flag through the admin `/wallets` handler and hide the provider from GET results while the feature is disabled.
  - Guard the PATCH endpoint so the embedded wallet toggle can’t be turned back on unless EnableEmbeddedWallets is true.

### Why

SDP Embedded Wallet wallet provider should not be available if embedded wallet feature is not enabled

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/wallets_handler_test.go
+++ b/internal/serve/httphandler/wallets_handler_test.go
@@ -342,11 +342,6 @@ func createWalletAssetsTestSetup(t *testing.T, ctx context.Context, dbPool db.SQ
 	}
 }
 
-func ptrBool(v bool) *bool {
-	b := v
-	return &b
-}
-
 func Test_WalletsHandlerPostWallets(t *testing.T) {
 	dbConnectionPool := getDBConnectionPool(t)
 


internal/serve/httphandler/wallets_handler_test.go
```diff
@@ -248,6 +270,16 @@ func Test_WalletsHandlerGetWallets(t *testing.T) {
...
+			prevFlag := handler.EnableEmbeddedWallets
+			if tc.enableEmbedded != nil {
+				handler.EnableEmbeddedWallets = *tc.enableEmbedded
+				require.Equal(t, *tc.enableEmbedded, handler.EnableEmbeddedWallets)
+			} else {
+				handler.EnableEmbeddedWallets = true
+			}
+			t.Cleanup(func() {
+				handler.EnableEmbeddedWallets = prevFlag
+			})
```
philipliu: What do you think about moving the handler initialization here so that we don't have to manage state between test cases?

philipliu (approved):
Minor comments. LGTM

PR #943, opened 2025-11-11
Fix asset code case sensitivity - Stellar asset codes are case-sensitive but we were converting all codes to uppercase causing issues with trust lines

Alonzo:
### What
Fixed asset code case sensitivity in the CreateAsset handler. Previously, all asset codes were converted to uppercase using 'strings.ToUpper()', but Stellar asset codes are case-sensitive. This caused issues when users entered assets with lowercase letters leading to trustline mismatches. The fix preserves the exact case entered by users while still using case-insensitive comparison for native XLM asset validation.

### Why
Stellar asset codes are case-sensitive identifiers. When users create assets with mixed case (like 'USDM0a'), the system was converting them to uppercase ('USDM0A'), which could cause issues when the actual asset on-chain has a different case or users try to add trustlines that don't match the stored case.

### Known limitations
N/A

### Checklist
- [X] Title follows `SDP-1234: Add new feature` or `Chore: Refactor package xyz` format. The Jira ticket code was included if available.
- [X] PR has a focused scope and doesn't mix features with refactoring
- [X] Preview deployment works as expected
- [X] Ready for production


philipliu:
Thanks for the fix. There are a few more places where asset codes were still case-insensitive. I've addressed it here https://github.com/stellar/stellar-disbursement-platform-backend/pull/967

PR #944, opened 2025-11-11
SDP-1916: Improve SEP-10 middleware error handling

lijamie98:
### What

- Improved error message separation in SEP-10 middleware (missing vs invalid authorization header)
- Added special handling for expired tokens: returns 400 BadRequest instead of 401 Unauthorized
- Updated tests to reflect the new error handling behavior

### Why

- Better error messages help clients diagnose authentication issues
- Expired tokens are a client error (400) rather than an authentication failure (401)
- Aligns with HTTP status code semantics: 401 for invalid/missing credentials, 400 for expired tokens

Refer to: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#error

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/sepauth/sep10_auth_middleware_test.go
+++ b/internal/sepauth/sep10_auth_middleware_test.go
@@ -173,7 +173,8 @@ func Test_SEP10HeaderTokenAuthenticateMiddleware(t *testing.T) {
 			nextHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				claimsFromContext = GetSEP10Claims(r.Context())
 				w.WriteHeader(http.StatusOK)
-				w.Write([]byte("success")) //nolint:err check
+				_, err := w.Write([]byte("success"))
+				require.NoError(t, err)
 			})
 
 			handler := middleware(nextHandler)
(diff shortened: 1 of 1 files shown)


internal/sepauth/sep10_auth_middleware.go
```diff
@@ -36,7 +36,7 @@ func SEP10HeaderTokenAuthenticateMiddleware(jwtManager *JWTManager) func(http.Ha
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 			sep10Claims, err := jwtManager.ParseSEP10TokenClaims(token)
 			if err != nil {
-				httperror.Unauthorized("Invalid token", err, nil).Render(rw)
```
JiahuiWho: I think it should stay as 401.

When SEP‑10 token parsing fails and we don’t know who the caller is, 401 Unauthorized should be returned to signal “provide a valid bearer token.” Using 403 would imply the user is already authenticated but lacks permission, which isn’t the case here.

lijamie98: Added more detailed handling. When the token is not expired and invalid, we will return `Unauthorized`.

marwen-abid (approved):

PR #946, opened 2025-11-14
SDP-1863: Implement SEP-45 challenge validation

philipliu:
### What

This implements the SEP-45 challenge validation flow https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0045.md#token

### Why

The Anchor Platform has been removed, so we need to reimplement SEP-45.

### Known limitations

The HTTP handler, dependency injection, and JWT generation will be implemented separately to keep the size of this PR reasonable.

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/services/sep45_service.go
+++ b/internal/services/sep45_service.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net/url"
+	"slices"
 	"strings"
 
 	"github.com/stellar/go/clients/stellartoml"
@@ -153,7 +154,7 @@ func (s *sep45Service) CreateChallenge(ctx context.Context, req SEP45ChallengeRe
 	}
 
 	if !s.isValidHomeDomain(homeDomain) {
-		return nil, fmt.Errorf("invalid home_domain must match %s", s.getBaseDomain())
+		return nil, fmt.Errorf("home_domain must match %s", s.getBaseDomain())
 	}
 
 	clientDomain := ""
@@ -275,6 +276,7 @@ func (s *sep45Service) CreateChallenge(ctx context.Context, req SEP45ChallengeRe
 
 type webAuthVerifyArgs struct {
 	raw                 map[string]string // Useful for comparing arguments
+	xdr                 xdr.ScVec         // Original XDR arguments
 	clientAccount       string
 	clientContractID    xdr.ContractId
 	homeDomain          string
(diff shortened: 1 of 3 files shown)


internal/services/sep45_service.go
```diff
@@ -272,8 +273,160 @@ func (s *sep45Service) CreateChallenge(ctx context.Context, req SEP45ChallengeRe
...
+		clientDomainFound   bool
+	)
+
+	for _, entry := range entries {
+		contractFn, err := s.ensureWebAuthInvocation(entry)
+		if err != nil {
+			return nil, err
+		}
+
+		// Extract the invocation arguments and make sure they are valid and consistent across entries
```
marwen-abid: nit: the following logic can be extracted to a method to simplify the `ValidateChallenge` method. 
```
// Extract the invocation arguments and make sure they are valid and consistent across entries
if parsedArgs, err = s.validateArguments(contractFn.Args, parsedArgs, webAuthDomain); err != nil {
	return nil, fmt.Errorf("validating invocation arguments: %w", err)
}
```

internal/services/sep45_service.go
```diff
@@ -272,8 +273,160 @@ func (s *sep45Service) CreateChallenge(ctx context.Context, req SEP45ChallengeRe
...
+			argsXDR = contractFn.Args
+			parsedArgs, err = s.buildChallengeArgs(argsMap, webAuthDomain)
+			if err != nil {
+				return nil, err
+			}
+		} else if err := compareArgs(argsMap, parsedArgs.raw); err != nil {
+			return nil, err
+		}
+
+		// Check that we have the expected authorization entries
```
marwen-abid: nit: same for the following. Can be extracted into its own method. An idea is to have a separate struct `authEntryTracker` 

```go
// authEntryTracker tracks which required authorization entries have been found during validation.
type authEntryTracker struct {
	serverVerified    bool
	clientFound       bool
	clientDomainFound bool
}

// validate ensures all required authorization entries are present.
func (t *authEntryTracker) validate(requireClientDomain bool) error {
	if !t.serverVerified {
		return fmt.Errorf("missing signed server authorization entry")
	}
	if !t.clientFound {
		return fmt.Errorf("missing client account authorization entry")
	}
	if requireClientDomain && !t.clientDomainFound {
		return fmt.Errorf("missing client domain authorization entry")
	}
	return nil
}

// processEntry validates and classifies an authorization entry, updating the tracker state.
func (t *authEntryTracker) processEntry(entry xdr.SorobanAuthorizationEntry, parsedArgs *webAuthVerifyArgs, svc *sep45Service) error { 
...
}
```

PR #964, opened 2025-12-02
chore: fix flaky assets handler test due to time drift

marwen-abid:
### What
Updated the test  `Test_AssetHandler_submitChangeTrustTransaction_makeSurePreconditionsAreSetAsExpected` to use a fresh `txnbuild.Preconditions` instance for its expectations instead of relying on the package-level `defaultPreconditions` variable.

### Why
The `defaultPreconditions` variable is initialized when the package loads. When running the entire test suite, the time elapsed between package initialization and the execution of this specific test exceeds the allowed time drift (30s). This caused the test to fail because the transaction created inside the handler uses a fresh timestamp (time.Now()), while the test expectation was comparing it against the stale timestamp from `defaultPreconditions`. Using a fresh timestamp in the test expectation aligns it with the handler's behavior and prevents the time drift error.

### Checklist

- [ ] Ready for production


JiahuiWho (approved):
LGTM

PR #984, opened 2025-12-10
SDP-1923: Implement supported assets endpoint

philipliu:
### What

This implements the supported assets endpoint, which the embedded wallet will call to fetch balances.

### Why

Trustlines do not exist for contract accounts, so we need to tell the frontend which assets to fetch the balances for.

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/serve.go
+++ b/internal/serve/serve.go
@@ -741,16 +741,17 @@ func handleHTTP(o ServeOptions) *chi.Mux {
 					r.Post("/", walletCreationHandler.CreateWallet)
 					r.Get("/{credentialID}", walletCreationHandler.GetWallet)
 
-					// Wallet profile routes
-					r.With(middleware.WalletAuthMiddleware(o.walletJWTManager)).Route("/profile", func(r chi.Router) {
-						r.Get("/", embeddedWalletProfileHandler.GetProfile)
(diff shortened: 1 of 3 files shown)


internal/serve/serve.go
```diff
@@ -741,7 +742,10 @@ func handleHTTP(o ServeOptions) *chi.Mux {
 					r.Get("/{credentialID}", walletCreationHandler.GetWallet)
 
 					// Wallet profile routes
-					r.With(middleware.WalletAuthMiddleware(o.walletJWTManager)).Get("/profile", embeddedWalletProfileHandler.GetProfile)
+					r.With(middleware.WalletAuthMiddleware(o.walletJWTManager)).Route("/profile", func(r chi.Router) {
+						r.Get("/", embeddedWalletProfileHandler.GetProfile)
+						r.Get("/assets", embeddedWalletProfileHandler.GetAssets)
```
marwen-abid: We usually add all endpoints to `serve_test.go`. Like we have `Test_handleHTTP_unauthenticatedEndpoints` and `Test_handleHTTP_authenticatedEndpoints` . Not sure where this falls given that it's a new type of authentication but may be worth adding something there for the sake of consistency.

PR #985, opened 2025-12-11
SDP-1922: Consolidate wallet data retrieval under `/profile` endpoint

JiahuiWho:
### What

Consolidated the `embedded_wallet_profile_handler` so it fetches verification state, supported assets, and receiver contact, with a shared helper for translating service errors into 401/500 responses.

### Why

we can gate everything off the `isPending` flag,  and only fetch wallet details when verification already passed. This keeps the response aligned with the flow the client has to follow, and avoids partial states.

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/embedded_wallet_profile_handler.go
+++ b/internal/serve/httphandler/embedded_wallet_profile_handler.go
@@ -61,14 +61,28 @@ func (h EmbeddedWalletProfileHandler) GetProfile(rw http.ResponseWriter, req *ht
 	}
 
 	isPending, err := h.EmbeddedWalletService.IsVerificationPending(ctx, contractAddress)
-	if renderWalletServiceError(ctx, rw, err, "Failed to evaluate verification requirement") {
+	if err != nil {
+		switch {
+		case errors.Is(err, services.ErrMissingContractAddress):
+			httperror.Unauthorized("", err, nil).Render(rw)
+		case errors.Is(err, data.ErrRecordNotFound):
+			httperror.NotFound("receiver wallet not found", err, nil).Render(rw)
+		default:
+			httperror.InternalError(ctx, "Failed to evaluate verification requirement", err, nil).Render(rw)
+		}
 		return
 	}
 
 	var pendingAsset *data.Asset
 	if isPending {
 		pendingAsset, err = h.EmbeddedWalletService.GetPendingDisbursementAsset(ctx, contractAddress)
-		if renderWalletServiceError(ctx, rw, err, "Failed to retrieve pending disbursement asset") {
+		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/serve/httphandler/embedded_wallet_profile_handler_test.go
```diff
@@ -93,6 +93,98 @@ func Test_EmbeddedWalletProfileHandler_GetProfile(t *testing.T) {
...
+				var body EmbeddedWalletReceiverContactResponse
+				require.NoError(t, json.Unmarshal(resp.Body.Bytes(), &body))
+				require.NotNil(t, body.ReceiverContact)
+				assert.Equal(t, ttc.expected.Type, body.ReceiverContact.Type)
+				assert.Equal(t, ttc.expected.Value, body.ReceiverContact.Value)
+			})
+		}
+	})
+
+	t.Run("unauthorized when receiver contact missing", func(t *testing.T) {
```
philipliu: We should add a test for an empty contact address in the context.

internal/serve/serve_test.go
```diff
@@ -706,6 +706,7 @@ func Test_handleHTTP_authenticatedEndpoints(t *testing.T) {
 		// Embedded wallets
 		{http.MethodGet, "/embedded-wallets/profile"},
 		{http.MethodGet, "/embedded-wallets/profile-assets"},
+		{http.MethodGet, "/embedded-wallets/profile-receiver"},
```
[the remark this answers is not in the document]
JiahuiWho: Honest I’d like to merge the three profile endpoints into a single to keeps the wallet metadata in sync and saves clients multiple round-trips.

internal/serve/httphandler/embedded_wallet_profile_handler.go
```diff
@@ -89,8 +124,36 @@ func (h EmbeddedWalletProfileHandler) GetAssets(rw http.ResponseWriter, req *htt
...
+	if err == nil {
+		return false
+	}
+
+	switch {
+	case errors.Is(err, services.ErrMissingContractAddress), errors.Is(err, data.ErrRecordNotFound):
+		httperror.Unauthorized("", err, nil).Render(rw)
+	default:
+		httperror.InternalError(ctx, internalErrMsg, err, nil).Render(rw)
+	}
```
philipliu: Can we inline the error handling here? I’m not sure every call this wrapper covers returns `ErrRecordNotFound`, and even if it does, I don’t think we should always map that to `Unauthorized`. In some cases, it probably should be like a `NotFound` (or a different error) depending on context.

philipliu (approved):

PR #1015, opened 2026-01-19
SDP-1981: Validate tenant and account in transaction sponsorship

philipliu:
### What

This change adds `tenant_id` to the wallet JWT so that embedded wallet JWTs cannot be used across tenants. This also adds account validation to the sponsored transaction endpoints, so that any authenticated wallet user can fetch the transaction status for any other transaction if the ID is known.

### Why

Fixes two security bugs in the embedded wallet APIs.

### Known limitations

NA

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/passkey_handler.go
+++ b/internal/serve/httphandler/passkey_handler.go
@@ -97,17 +97,18 @@ func (h PasskeyHandler) FinishPasskeyRegistration(rw http.ResponseWriter, req *h
 
 	credential, err := h.WebAuthnService.FinishPasskeyRegistration(ctx, token, req)
 	if err != nil {
-		if errors.Is(err, wallet.ErrInvalidToken) {
+		switch {
+		case errors.Is(err, wallet.ErrInvalidToken):
 			httperror.BadRequest("Invalid token", err, nil).Render(rw)
-		} else if errors.Is(err, wallet.ErrWalletAlreadyExists) {
+		case errors.Is(err, wallet.ErrWalletAlreadyExists):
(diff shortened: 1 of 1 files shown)


internal/serve/httphandler/passkey_handler.go
```diff
@@ -215,20 +226,32 @@ func (h PasskeyHandler) RefreshToken(rw http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	credentialID, contractAddress, err := h.WalletJWTManager.ValidateToken(ctx, reqBody.Token)
+	credentialID, contractAddress, tokenTenantID, err := h.WalletJWTManager.ValidateToken(ctx, reqBody.Token)
 	if err != nil {
 		if errors.Is(err, wallet.ErrExpiredWalletToken) {
```
marwen-abid: nit: probably more idiomatic to write it as follows : 
```go
if err != nil {
		switch {
		case errors.Is(err, wallet.ErrExpiredWalletToken):
			httperror.Unauthorized("Token has expired", err, nil).Render(rw)
		case errors.Is(err, wallet.ErrInvalidWalletToken):
			httperror.Unauthorized("Invalid token", err, nil).Render(rw)
		case errors.Is(err, wallet.ErrMissingSubClaim),
			errors.Is(err, wallet.ErrMissingTenantClaim):
			httperror.Unauthorized("Invalid token claims", err, nil).Render(rw)
		default:
			httperror.InternalError(ctx, "Failed to validate token", err, nil).Render(rw)
		}
		return
	}
```

marwen-abid (approved):
LGTM!

PR #1025, opened 2026-01-23
SDP-2003 Reuse existing pending embedded wallet on invitation resend

marwen-abid:
### What                                                                                                                                                                                                                                
Reuse invite link when there is already an existing pending embedded wallet for the receiver. 

### Why                                                                                                                                                                                                                                     

 When resending an invitation for an embedded wallet via `PATCH /receivers/wallets/{id}`, the system was creating fresh entries in the `embedded_wallets` table for each resend attempt, even when a PENDING registration already        
     existed for that receiver wallet. This caused:                                                                                                                                                                                      
                                                                                                                                                                                                                                         
     1. Duplicate entries accumulating in the embedded_wallets table                                                                                                                                                                     
     2. Previously sent invitation links becoming orphaned/invalid                                                                                                                                                                       
     3. Potential confusion for receivers who might have multiple valid-looking registration tokens     

### Checklist

- [ ] Ready for production


philipliu (commented):
This change makes sense. A receiver will now have only 1 active invite at a time, meaning that if the `None` verification type is selected, there is no ambiguity about where the funds should go. The only risk is that if a user's token is somehow leaked, their funds could be stolen, which is a risk of using the `None` verification type in the first place.

PR #1051, opened 2026-02-10
Fix: Fix short linking is not enabled by default

JiahuiWho:
### What

Enable `IsLinkShortenerEnabled` when setup tenant
This is fix to https://github.com/stellar/stellar-disbursement-platform-backend/pull/916

#### Why my previous change did not work
1. Migration `2023-03-16.0`: Creates the `organizations` table and inserts a row — the org row now exists
2. Migration `2025-01-30.0`: Adds `is_link_shortener_enabled` column with DEFAULT false — since the row already exists from step 1, PostgreSQL backfills it with false
3. Migration `2025-09-20.0` (my previous PR): Changes the default to true — but the row already exists with false from step 2, so SET DEFAULT has no effect

The `SET DEFAULT` only applies to future `INSERT` statements that omit the column. But the org row was already inserted in step 1, and the column was backfilled in step 2. By the time my migration runs, there's nothing left to insert — the damage is already done.

### Known limitations

Since that migration was already merged and deployed, deleting it could cause the migration tool to get confused about its state on those environments. I'd just leave it in place, it's a no-op anyway. 

### Checklist

- [ ] Ready for production

### DIFF
--- a/stellar-multitenant/internal/provisioning/manager.go
+++ b/stellar-multitenant/internal/provisioning/manager.go
@@ -304,12 +304,11 @@ func (m *Manager) setupTenantData(ctx context.Context, tenantSchemaDSN string, p
 		return fmt.Errorf("running setup wallets for proper network: %w", err)
 	}
 
-	isLinkShortenerEnabled := true
 	err = models.Organizations.Update(ctx, &data.OrganizationUpdate{
 		Name:                   pt.OrgName,
 		MFADisabled:            pt.MFADisabled,
 		CAPTCHADisabled:        pt.CAPTCHADisabled,
-		IsLinkShortenerEnabled: &isLinkShortenerEnabled,
+		IsLinkShortenerEnabled: utils.Ptr(true),
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


stellar-multitenant/internal/provisioning/manager.go
```diff
@@ -304,10 +304,12 @@ func (m *Manager) setupTenantData(ctx context.Context, tenantSchemaDSN string, p
...
 
+	isLinkShortenerEnabled := true
 	err = models.Organizations.Update(ctx, &data.OrganizationUpdate{
-		Name:            pt.OrgName,
-		MFADisabled:     pt.MFADisabled,
-		CAPTCHADisabled: pt.CAPTCHADisabled,
+		Name:                   pt.OrgName,
+		MFADisabled:            pt.MFADisabled,
+		CAPTCHADisabled:        pt.CAPTCHADisabled,
+		IsLinkShortenerEnabled: &isLinkShortenerEnabled,
```
philipliu: nit: We can use `utils.Ptr(true)` instead

PR #1052, opened 2026-02-11
SDP-2009: Add endpoint for fetch captcha config

JiahuiWho:
### What

1. Add `/app-config` endpoint to fetch captcha config for FE use
2. Change`forgot_password_handler.go` to check `IsCAPTCHADisabled()` instead of env var `h.ReCAPTCHADisabled`, which is only the  env-level flag.

### Why

reCAPTCHA v3

### Known limitations

To test reCAPTCHA v3 locally, you will need to update `.env` with
```
CAPTCHA_TYPE="GOOGLE_RECAPTCHA_V3"
RECAPTCHA_SITE_KEY="6LcgOWcsAAAAAJkeyQHEW4-yA-Eu79kYHDYXeznF"
RECAPTCHA_SITE_SECRET_KEY="6LcgOWcsAAAAACGUHDBr7VpcZ_C5BUjRPuBk9EUP"
```
and `env-config-testnet.js` (the nginx)with
```
RECAPTCHA_SITE_KEY: "6LcgOWcsAAAAAJkeyQHEW4-yA-Eu79kYHDYXeznF",
```

### Checklist

- [ ] Ready for production


internal/serve/httphandler/captcha_config_handler.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (h CAPTCHAConfigHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	orgName := r.URL.Query().Get("organization_name")
+
+	var tnt *schema.Tenant
+	var err error
+
+	if h.SingleTenantMode {
```
marwen-abid: If you make this endpoint tenant aware using `r.Use(middleware.EnsureTenantMiddleware)`, you wouldn't need to resolve the tenant manually (see comment below). 

You can just use `t, err := GetTenantFromContext(ctx)`

internal/serve/httphandler/captcha_config_handler.go
```diff
@@ -0,0 +1,68 @@
+package httphandler
```
marwen-abid: 🚨  I believe having an endpoint just for `captcha-config` is overkill. But also there are other properties we'll want to add to this endpoint later. 

The ultimate goal will be to remove `env-config.js` from the FE and dynamically load all properties through this endpoint. So things like : 
```
STELLAR_EXPERT_URL, HORIZON_URL, RPC_ENABLED, RECAPTCHA_SITE_KEY,  SINGLE_TENANT_MODE
```

All these non-sensitive, properties will be loaded through this endpoint. So i think we'll have to name it something else, more generic. Possible names : 
- `GET /info` 
- `GET /config` 
- `GET /app-config` 
- `GET /settings`

PR #1068, opened 2026-02-20
Fix: Check muxed account auth is not infra account in sponsored transaction worker

philipliu:
### What

This PR hardens the sponsored transaction auth validation by checking muxed account auth entries in addition to regular account auth entries. We now reject sponsored transactions that require auth from the channel or distribution infra accounts, even when those addresses are represented as muxed accounts. This is a defensive change as the sponsored transaction handler already rejects these types of transactions, which are first simulated in enforcing mode (requiring all signatures to be present and valid).

### Why

Additional hardening of TSS to prevent accidental insertion of bad sponsored transaction records.

### Known limitations

N/A

### Checklist

- [ ] Ready for production


marwen-abid (approved):
LGTM!

PR #1115, opened 2026-04-20
SDP-2073: Add ownership check to SEP-24 GET /transaction endpoint         

JiahuiWho:
### What

Add account ownership verification to `GET /sep24/transaction` so it only returns transaction details when the authenticated SEP-10/SEP-45 subject matches the receiver wallet on file, returns 404 when the caller does not own the requested transaction.

### Why

To align with the SEP-24 spec requirement
https://hackerone.com/reports/3681382

### Known limitations

N/A

### Checklist

- [ ] Ready for production

### DIFF
--- a/internal/serve/httphandler/sep24_handler.go
+++ b/internal/serve/httphandler/sep24_handler.go
@@ -136,7 +136,8 @@ func (h SEP24Handler) GetTransaction(w http.ResponseWriter, r *http.Request) {
 		"refunded": false, // Always false for registration
 	}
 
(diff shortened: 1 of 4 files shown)


internal/serve/httphandler/sep24_handler.go
marwen-abid: nit: We can change this method to be `GetBySEP24TransactionIDAndAccount(ctx, txID, account, memo)`, which does the lookup using the account and memo, and fold ownership + lookup into one step

marwen-abid (approved):
LGTM!

PR #1130, opened 2026-05-08
Harden Claude release workflow security

leighmcculloch:
### What

Apply a best practice to the `automated_release_process` workflow:

- Move the broad `contents: write`/`pull-requests: write`/`issues: write`/`id-token: write` scope from workflow level to the single job that needs it; set top-level `permissions: {}` so any future job added to this file inherits nothing by default.

### Why

This is pretty minor, but brings this workflow in line with the Claude-action security standards being applied across the stellar org (see `stellar/stellar-core` [PR #5258](https://github.com/stellar/stellar-core/pull/5258) and the [claude-code-action security guide](https://github.com/anthropics/claude-code-action/blob/main/docs/security.md)). Without `permissions: {}` at the workflow root, any future job added to this file would silently inherit the broad write scopes. Moving them to the job level limits the blast radius if this workflow expands in the future.

### DIFF
--- a/.github/workflows/automated_release_process.yml
+++ b/.github/workflows/automated_release_process.yml
@@ -1,10 +1,9 @@
 name: Automated Release Process
 
-permissions:
-  contents: write
-  pull-requests: write
-  issues: write
-  id-token: write
+# Revoke all default GITHUB_TOKEN permissions at the workflow level. Each job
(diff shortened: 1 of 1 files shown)


marwen-abid (approved):
LGTM! Thank you for this 🙏

PR #1119, opened 2026-04-24
SDP-2065: Add option to disable receiver wallet invitations per tenant

marwen-abid:
## What

Adds a new `receiver_invitations_disabled` boolean setting on the organization profile. When set to `true`, the scheduled receiver wallet invitation job skips sending invitations for that tenant. 

The flag is readable via `GET /profile/organization` and updatable via `PATCH /profile/organization`, and is backed by a new nullable column on the `organizations` table.

## Why

Tenants need a way to pause outbound receiver wallet invitations.

### Checklist

- [x] Title follows `SDP-1234: Add new feature` or `Chore: Refactor package xyz` format. The Jira ticket code was included if available.
- [x] PR has a focused scope and doesn't mix features with refactoring
- [x] Tests are included (if applicable)
- [x] `CHANGELOG.md` is updated (if applicable)
- [ ] If contracts changed, run the `Contract WASM Artifacts` workflow and open a PR to update the WASMs on `dev`
- [ ] CONFIG/SECRETS changes are updated in helmcharts and deployments (if applicable)
- [ ] Preview deployment works as expected
- [ ] Ready for production

PR #1161, opened 2026-07-01
feat: multi-distribution-wallet support

xBalbinus:
## Multi-distribution-wallet support

Lets a tenant configure and operate multiple distribution wallets with per-wallet authorized users, so a tenant's on-chain activity across programs is no longer linkable to one address, organizational/compliance boundaries are cleaner, and a key compromise is contained to a single wallet. **Single-wallet behavior stays the default** for existing and new tenants — the feature is opt-in.

### Wallet management
Owner can create, name, and describe additional distribution wallets, each independently provisioned + funded with its own separately-encrypted secret material (envelope encryption: per-wallet DEK wrapped by the host passphrase, so rotating one wallet's key never touches another's ciphertext). Wallets can be archived (no new disbursements) but never hard-deleted once they've submitted transactions, preserving audit history. Atomic promote-to-default; zero-active-wallets invariant enforced at API + DB. *v1 note:* additional wallets are `DB_VAULT`-type (mixed-custody guard); the account-type-per-wallet generalization is a follow-up. 20-wallet cap per tenant in v1.

### Wallet-scoped authorization
Adds `(user, wallet, role)` membership alongside the existing `(user, tenant, role)` records. Tenant Owner stays org-wide; every other role (Financial Controller, Developer, Initiator, Approver) can be scoped to specific wallets. Action authorization is `(role, wallet)`-gated — an Approver on Wallet A cannot approve a disbursement targeting Wallet B. Roles are unchanged; this is an added scoping dimension.

### Context switching without re-auth
One login surfaces every wallet the user is entitled to and their per-wallet role; the operator switches wallet context without re-authenticating. Per-request operations carry the active wallet in the `X-Wallet-Id` header; the server validates the caller's membership on that wallet. *Design note vs the brief:* entitlements are resolved per-request from membership records rather than encoded in the JWT — this keeps grants/revocations effective immediately and avoids JWT-size limits, at the cost of one indexed lookup per request.

### Disbursement routing
Each disbursement is routed to an explicit source wallet via `X-Wallet-Id` (a multi-wallet tenant gets a `400` if it's absent — no silent default); unentitled or mismatched wallets are rejected at every state transition; `source_wallet_id` is a `NOT NULL` FK with `ON DELETE RESTRICT` and is immutable after creation. Single-wallet tenants keep the implicit default. *v1 note:* rule-based routing (by program/country/asset) is a follow-up; v1 is explicit-at-creation.

### Observability & audit
`wallet_id` tags on records, structured logs, and metrics; wallet-aware statistics + per-wallet balance + owner-only all-wallets aggregate; wallet-scoped receivers and CSV exports; immutable wallet-membership audit trail (append-only reject trigger); transactional webhook outbox.

### Backward compatibility & cutover
- Existing tenants get an implicit **default** wallet that mirrors their current distribution account (address/type/status copied from `admin.tenants`), one per tenant schema. Single-wallet behavior is unchanged unless a tenant opts in.
- Existing disbursements and payments are backfilled to that default wallet (disbursement payments inherit their disbursement's wallet; direct payments → default).
- **No key migration:** the default (legacy) account's secret stays in the existing `vault`; signing falls back to the vault for it, while new wallets use per-wallet envelope-encrypted keys (precedence + fallback covered by test).
- **Cutover playbook:** `docs/multi-wallet/CUTOVER.md` (+ `docs/multi-wallet/preflight.sql`) covers the pre-flight checks, the online-safe DDL for large tenants, and post-migration verification. The `.2`/`.6` backfills now **fail fast** with a clear error if a schema has data but no default wallet (instead of an opaque NOT NULL abort).
- **Two things to run before cutover:** a pre-flight that flags (a) any tenant schema with data but no default wallet and (b) any non-owner user whose roles are outside the wallet-scopable set (they'd get zero memberships and be locked out); and a heads-up to API integrators that **`X-Wallet-Id` becomes required on writes once a tenant has a 2nd active wallet** (a data-triggered behavior change, not a deploy one).
- **Migration online-safety is the main operational item:** the `source_wallet_id` backfill + `SET NOT NULL` + FK + index on `disbursements`/`payments` take write-blocking locks; large tenants need the online-safe DDL rewrite (batched backfill, `CHECK NOT VALID`+`VALIDATE`, `CREATE INDEX CONCURRENTLY`) and a per-tenant pre-flight (every tenant schema with disbursements must have a matching non-deleted tenant so the default wallet exists, or `SET NOT NULL` aborts).

### Testing
6,400+ tests + verification specs. Verified end-to-end on testnet: provisioning, envelope-encryption isolation, wallet-scoped authz, `X-Wallet-Id` routing, and on-chain USDC + XLM disbursements settling PENDING→SUCCESS. Includes a regression test for a TSS metric label-cardinality crash found during end-to-end testing.

### Known follow-ups
- Migration online-safety rewrite for large tenants (above).
- Outbound webhook HMAC signing + `webhook_url` scheme/SSRF validation.
- `is_approval_required` (four-eyes) defaults off.

Opened as **draft** pending SDF coordination on the migration window + secrets handling.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,26 @@
+# Continuous deployment for the multi-wallet PLAYTEST demo (Railway project sdp-mw-playtest).
+# Railway's own push-to-deploy was not firing, so we drive it from Actions: every push to
+# feat/multi-wallet runs `railway up` against the sdp-api service. RAILWAY_TOKEN is a
+# project-scoped token (repo secret), so the CLI needs no `railway link`.
+name: Deploy sdp-api to Railway (playtest)
```
hypekostas: Demo-specific, should be removed before merging.

railway.json
```diff
@@ -0,0 +1,11 @@
+{
```
hypekostas: Demo-specific, should be removed before merging.

internal/data/payments.go
```diff
@@ -663,6 +700,11 @@ func newPaymentQuery(baseQuery string, queryParams *QueryParams, sqlExec db.SQLE
 	if queryParams.Filters[FilterKeyStatus] != nil {
 		addArrayOrSingleCondition[PaymentStatus](qb, "p.status", queryParams.Filters[FilterKeyStatus])
 	}
+	if walletIDs, ok := queryParams.Filters[FilterKeySourceWalletIDs].([]string); ok {
+		// Membership-filtered visibility. Direct payments (no disbursement) carry no
+		// explicit source wallet until and inherit default-wallet visibility.
+		qb.AddCondition("COALESCE(d.source_wallet_id, (SELECT dw.id FROM distribution_wallets dw WHERE dw.is_default)) = ANY(?)", pq.Array(walletIDs))
```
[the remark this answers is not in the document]
hypekostas: Confirmed this needs addressing both here and in payments_handler.go:92-109.
The comment seems to imply this was deliberate but I assume it was leftover from pre-W3.

internal/serve/httphandler/payments_handler.go
```diff
@@ -87,6 +87,33 @@ func (p PaymentsHandler) GetPayment(w http.ResponseWriter, r *http.Request) {
...
+			if payment.Disbursement != nil {
+				sourceWalletID = payment.Disbursement.SourceWalletID
+			}
+			if sourceWalletID == "" {
+				defaultWallet, dwErr := p.Models.DistributionWallets.GetDefault(ctx, p.DBConnectionPool)
+				if dwErr != nil {
+					httperror.InternalError(ctx, "Cannot resolve wallet visibility", dwErr, nil).Render(w)
+					return
+				}
+				sourceWalletID = defaultWallet.ID
```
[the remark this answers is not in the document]
hypekostas: Confirmed this needs addressing both here and in payments.go:706. 
Please also update the comment at line 92, assuming it was leftover from pre-W3.
Copilot's cross-reference to line 395 of this file looks to be wrong.

internal/serve/httphandler/payments_handler.go
```diff
@@ -319,6 +391,15 @@ func (p PaymentsHandler) PostDirectPayment(w http.ResponseWriter, r *http.Reques
...
+	//  routing: explicit source wallet via X-Wallet-Id (single-wallet tenants may omit).
+	sourceWallet, walletErr := resolveSourceWalletForWrite(ctx, r, p.AuthManager, p.Models,
+		data.FinancialControllerUserRole, data.BusinessUserRole)
+	if walletErr != nil {
+		walletErr.Render(w)
+		return
+	}
+	serviceReq.SourceWalletID = sourceWallet.ID
+
 	payment, err := p.DirectPaymentService.CreateDirectPayment(ctx, serviceReq, user, &distAccount)
```
hypekostas, line 399: Copilot's suppressed comment (low confidence) is correct: 

"Although this resolves the selected wallet, `CreateDirectPayment` is still passed the tenant's legacy `distAccount`. Its trustline and balance checks therefore run against the default account, and its pending-payment total is not scoped by wallet. Secondary-wallet payments can be incorrectly accepted or rejected; build the transaction account from `sourceWallet` and scope pending amounts to `SourceWalletID`."

This is already handled correctly in `disbursement_handler.go:659-673`.

internal/services/direct_payment_service.go
hypekostas: Similar issue to [this comment](https://github.com/stellar/stellar-disbursement-platform-backend/pull/1161/changes#r3665620557). 

Consider fix: thread the wallet ID into the balance check so it can scope the pending total:

`if err = s.validateBalance(ctx, dbTx, distributionAccount, asset, req.Amount, req.SourceWalletID); err != nil {`

Then add the parameter to `validateBalance` at Line 326 and pass it through at Line 367:

`totalPending, err := s.calculatePendingAmountForAsset(ctx, dbTx, *asset, sourceWalletID)`

hypekostas, line 395: In continuation of [previous comment](https://github.com/stellar/stellar-disbursement-platform-backend/pull/1161/changes#r3665716617):

Scope the pending query, matching the implementation in `disbursement_management_service.go:326-32`.

internal/transactionsubmission/transaction_worker.go
```diff
@@ -472,6 +496,12 @@ func (tw *TransactionWorker) prepareForSubmission(ctx context.Context, txJob *Tx
 }
 
 func (tw *TransactionWorker) buildAndSignTransaction(ctx context.Context, txJob *TxJob) (*txnbuild.FeeBumpTransaction, error) {
+	signStart := time.Now()
+	defer func() {
+		log.Ctx(ctx).WithField("duration_ms", time.Since(signStart).Milliseconds()).
+			Debug("built and signed fee-bump transaction")
+	}()
+
 	distributionAccount, err := tw.engine.DistributionAccountResolver.DistributionAccount(ctx, txJob.Transaction.TenantID)
```
hypekostas: See [this comment by copilot](https://github.com/stellar/stellar-disbursement-platform-backend/pull/1161#discussion_r3639996612) where the symptom originates.

TSS ignores `source_wallet_id` and every payment is sent from the default wallet (source account is resolved from the tenant record (`admin.tenants.distribution_account_address`), which `syncTenantDefaultAccount` keeps pointed at the default wallet.)

Fix: Make TSS account resolution wallet-aware. Resolve the distribution account from `txJob.Transaction.WalletID` when present (falling back to the tenant account for legacy rows and non-payment transaction types), and use that account as operation source, fee account, and signer. The signature client already resolves per-wallet keys by public key (`distribution_account_db_vault_signature_client.go:92`), so once the correct address is threaded through, signing works without further change.

internal/services/paymentdispatchers/stellar_payment_dispatcher.go
```diff
@@ -85,6 +86,7 @@ func (s *StellarPaymentDispatcher) sendPaymentsToTSS(ctx context.Context, sdpDBT
 				MemoType:    memo.Type,
 			},
 			TenantID: tenantID,
+			WalletID: sdpUtils.SQLNullString(payment.SourceWalletID),
```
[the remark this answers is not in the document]
hypekostas: This is where the symptom originates, but the fix site is in `transaction_worker.go` (see [this comment](https://github.com/stellar/stellar-disbursement-platform-backend/pull/1161/changes#r3665875124)).

internal/serve/httphandler/wallet_scope.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/stellar/stellar-disbursement-platform-backend/stellar-auth/pkg/auth"
+)
+
+// resolveWalletReadScope computes the caller's read-visibility scope for
+// membership-filtered endpoints (taxonomy): nil = Owner, no filtering; non-nil = the exact
+// wallet set the caller may see (possibly empty).
+func resolveWalletReadScope(ctx context.Context, authManager auth.AuthManager, models *data.Models) ([]string, *httperror.HTTPError) {
+	user, err := ctxHelper.GetUserFromContext(ctx, authManager)
+	if err != nil {
+		return nil, httperror.InternalError(ctx, "Cannot get user from context", err, nil)
```
hypekostas, line 27: Deactivating a user 500s every API key they created. API-key requests set the acting user to `apiKey.CreatedBy` (`api_keys_middleware.go:196`), so this call resolves the human who created the key. `GetUser` filters on `is_active = true` (`stellar-auth/pkg/auth/authenticator.go:469`), so once that person is deactivated (normal offboarding) this returns an error and we render a 500. Please confirm that this intended behavior.

The tests miss this because `serve_api_key_test.go:148,225` construct the acting user with `IsOwner: true`, short-circuiting before any membership lookup.

Fix: If not intended behavior, give API keys their own wallet scope rather than inheriting the creator's.

db/migrations/sdp-migrations/2026-06-09.5-backfill-wallet-memberships.sql
```diff
@@ -0,0 +1,39 @@
+-- Backfill wallet memberships: every existing non-owner user becomes a member of their
+-- tenant's DEFAULT wallet with their current role(s), so wallet-scoped authorization changes
+-- nothing for existing tenants (their only wallet is the default). Owner stays tenant-wide and
+-- receives no membership rows.
+--
+-- Guards make this safe in every layout:
+-- - fresh tenant schemas: sdp-migrations run before auth-migrations, so auth_users does not
+--   exist yet — no users to backfill (new users get memberships via the management API)
```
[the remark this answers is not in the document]
hypekostas: Confirmed, please address at `internal/serve/httphandler/user_handler.go:178` for both the single wallet and multi-wallet tenants. For single-wallet tenants especially, the failure is silent and the newly created user will see empty screens. Assuming this isn't intended.

Suggestions: either give the user a default wallet assignment upon creation that can be managed later by the owner, or (even better) allow wallet to be assigned at user creation (must also be addressed in the frontend), +  test for newly created user's wallet access.

cmd/root.go
```diff
@@ -10,19 +10,50 @@ import (
...
+// stdout capture entirely. This matters because Railway has no log-drain /
+// log-forwarding feature of its own (per Railway's docs), so without this the
+// structured fields the app already logs (payment_id, wallet_id, Horizon
+// error codes, ...) never reach Grafana/Loki in production.
+//
+// It's a no-op when globalOptions.LogShippingURL is empty, which is the
+// default -- local dev and any deployment that doesn't set LOG_SHIPPING_URL
+// are completely unaffected.
+func registerLogShippingHook() {
+	if globalOptions.LogShippingURL == "" || lokiHook != nil {
```
hypekostas, line 37: `globalOptions.LogShippingURL` can be set via CLI, but it would be useful if it were also added to helm charts values as an option.

cmd/root.go
```diff
@@ -10,19 +10,50 @@ import (
...
 var globalOptions cmdUtils.GlobalOptionsType
 
+// lokiHook is the process-wide Loki log-shipping hook, non-nil only when
+// LOG_SHIPPING_URL is set. It's populated once by registerLogShippingHook
+// (called from rootCmd's PersistentPreRun, right alongside where LOG_LEVEL is
+// applied) and read by the "serve" subcommand so it can be flushed on
+// graceful shutdown -- see ServeCommand.Command in cmd/serve.go.
+var lokiHook *observability.LokiHook
+
+// registerLogShippingHook wires the app's logs to also ship directly to a
```
hypekostas: Railway-specific workaround perhaps shouldn't survive in a comment. Consider generalizing, e.g.:

"Ships logs directly to a Loki-compatible push endpoint, for platforms where no log collector can read the container's stdout (PaaS environments without a log-drain feature). On Kubernetes, prefer a collector DaemonSet reading stdout."

internal/scheduler/jobs/event_delivery_job.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil // webhook delivery not configured for this tenant
+	}
+	webhookURL := *organization.WebhookURL
+
+	type eventRow struct {
+		ID      string `db:"id"`
+		Payload []byte `db:"payload"`
+	}
+	var pending []eventRow
+	err = j.models.DBConnectionPool.SelectContext(ctx, &pending, `
```
hypekostas, line 128: Low priority: This query at `event_delivery_job.go:70` doesn't reserve rows, so every running instance delivers the same events (only matters if running replica instances).

internal/data/organizations.go
```diff
@@ -55,8 +55,11 @@ type Organization struct {
 	MFADisabled                 *bool                  `json:"mfa_disabled" db:"mfa_disabled"`
 	CAPTCHADisabled             *bool                  `json:"captcha_disabled" db:"captcha_disabled"`
 	ReceiverInvitationsDisabled *bool                  `json:"receiver_invitations_disabled" db:"receiver_invitations_disabled"`
-	CreatedAt                   time.Time              `json:"created_at" db:"created_at"`
-	UpdatedAt                   time.Time              `json:"updated_at" db:"updated_at"`
+	// WebhookURL is the tenant-configured destination for outbox event delivery;
+	// delivery is skipped when unset.
+	WebhookURL *string   `json:"webhook_url,omitempty" db:"webhook_url"`
```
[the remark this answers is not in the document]
hypekostas: Indeed: the webhook outbox can't be enabled, and has some gaps once it can be.

`organizations.webhook_url` has no write path. `OrganizationUpdate` (`organizations.go:66-84`) doesn't include it and no handler sets it. The delivery job returns early on every run, so the outbox is dead code unless someone edits the DB by hand.

1. Give it a write path: Add `WebhookURL` to `OrganizationUpdate `and expose it on `PATCH /organization` (`profile_handler.go:160:165'), following how `PrivacyPolicyLink` is already handled.
2. Require https: Without it a tenant can configure http://, and every payment and disbursement event crosses the network in plaintext. No attacker needed for this one, it's just customer data in the clear. Same validation hook as above.
3. Sign the payload: Add a signature header alongside the POST in `deliver()` (`event_delivery_job.go:97`). Right now the URL is the only secret, and URLs leak into logs and proxies. Anyone who learns it can send fake payment events to operators, and they have no way to tell.
4. Restrict where deliveries can go: In `deliver()`, check that the host resolves to a public address before sending, and disable redirect-following. Otherwise a tenant admin can make SDP POST to internal services.

internal/services/distribution_wallet_management_service.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	dbPool := s.Models.DBConnectionPool
+
+	// Reserve the wallet row atomically. A per-tenant (per-schema) advisory lock makes the
+	// eligibility + cap checks and the insert one critical section: without it, two concurrent
+	// owner creates could each pass a stale Count check and exceed the cap (TOCTOU). The
+	// lock is held only for these fast DB ops and released at COMMIT — the on-chain keygen,
+	// funding, and trustline work below deliberately runs OUTSIDE this transaction.
+	wallet, err := db.RunInTransactionWithResult(ctx, dbPool, nil, func(dbTx db.DBTransaction) (*data.DistributionWallet, error) {
```
hypekostas, line 116: Lines 116-172: `CreateWallet` publishes a wallet before it's funded. 
The reserve transaction commits at `:145`. `Insert` omits both `status` and `distribution_account_address` (`internal/data/distribution_wallets.go:85-89`), so the row lands as `ACTIVE` with a `NULL` address, while keygen, `UpdateAddress`, and `CreateAndFundAccount` all still run afterwards at `:150-172`. In that window the wallet is indistinguishable from a working one, and disbursements and direct payments against it are accepted.

Suggested fix: mirror the tenant provisioning lifecycle. It uses a status that marks incompleteness (`pkg/schema/tenant.go:27-30`): `TENANT_CREATED` -> `TENANT_PROVISIONED`. A tenant that dies mid-provision is visibly unfinished; a wallet that dies mid-provision claims to be `ACTIVE`.

1. Add `PENDING` to the status `CHECK` (2026-06-09.0:22) and insert with it.
2. Promote to `ACTIVE` only after `CreateAndFundAccount` succeeds.
3. Treat non-`ACTIVE` like `ARCHIVED` in `GetAll(..., false)` and `resolveSourceWalletForWrite`.

Same shape as tenant provisioning, and it makes stranded rows self-evident instead of masquerading as usable.

db/migrations/sdp-migrations/2026-06-09.3-add-distribution-wallet-lifecycle-triggers.sql
```diff
@@ -0,0 +1,56 @@
...
+
+CREATE TRIGGER enforce_distribution_wallet_not_last_active_trigger
+    BEFORE UPDATE OF status ON distribution_wallets
+    FOR EACH ROW EXECUTE PROCEDURE enforce_distribution_wallet_not_last_active();
+
+-- +migrate StatementBegin
+CREATE OR REPLACE FUNCTION enforce_disbursement_source_wallet_active()
+RETURNS TRIGGER AS $$
+BEGIN
+    IF (SELECT status FROM distribution_wallets WHERE id = NEW.source_wallet_id) = 'ARCHIVED' THEN
```
hypekostas, line 37: `enforce_disbursement_source_wallet_active` (`2026-06-09.3:37`) and `enforce_wallet_membership_wallet_active` (`2026-06-09.4:45`) read `status` with an unlocked `SELECT`, so they see the pre-archive row for the full duration of an in-flight archive. 

Normal API traffic reaches it: `acquireTenantWalletLock` is taken only by `CreateWallet`, `ArchiveWallet` and `PromoteToDefault`, and both write paths do their ARCHIVED check outside the transaction that inserts (`disbursement_handler.go:174` vs `:196`; likewise `GrantMembership`).

With an archive open but uncommitted, both the disbursement insert and the membership grant succeed against a wallet that is ARCHIVED once it commits. Reverse the order and they reject correctly. The FK doesn't help , since `FOR KEY SHARE` doesn't conflict with a non-key `UPDATE` of `status`.

Fix: add `FOR SHARE` to the `SELECT` in both trigger functions so they lock the wallet row. It conflicts with the `FOR NO KEY UPDATE` the archive takes, so the two serialize in either order.

Related gap: there's no equivalent trigger on payments, so a direct payment can be created against an archived wallet with no database check whatsoever. The only guard is the app-level check in `resolveSourceWalletForWrite`, which runs outside the inserting transaction.

db/migrations/sdp-migrations/2026-06-09.6-add-source-wallet-id-to-payments.sql
```diff
@@ -0,0 +1,98 @@
...
+        ELSE
+            RAISE EXCEPTION 'direct payments must state their source distribution wallet explicitly'
+                USING ERRCODE = 'not_null_violation', CONSTRAINT = 'payments_source_wallet_required';
+        END IF;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+-- +migrate StatementEnd
+
```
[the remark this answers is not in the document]
hypekostas: Low severity, not a live bug.

The mechanism is right: `derive_payment_source_wallet` (`2026-06-09.6:47`) only derives when `source_wallet_id IS NULL`, so an explicit contradicting value would survive, permanently (both columns are immutable).

But no caller can supply one. There's no request field (`SourceWalletID` is `json:"-"`, set from `X-Wallet-Id`), and `InsertAll` (`payments.go:384`) omits the column entirely, so a value set on `PaymentInsert` is silently discarded rather than stored.

internal/transactionsubmission/transaction_worker.go
```diff
@@ -521,7 +551,13 @@ func (tw *TransactionWorker) buildAndSignTransaction(ctx context.Context, txJob
 }
 
 func (tw *TransactionWorker) submit(ctx context.Context, txJob *TxJob, feeBumpTx *txnbuild.FeeBumpTransaction) error {
+	log.Ctx(ctx).WithField("tx_hash", txJob.Transaction.StellarTransactionHash.String).
+		Info("submitted transaction to Horizon, awaiting result")
```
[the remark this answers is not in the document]
hypekostas: Indeed, present tense ("submitting…" instead of "submitted...") would fix the wording and keep the pairing with the `Debug` response line below.

Note this is submission failures only, build/signing failures return at STEP 2 (`prepareForSubmission`, `:422`) and never reach `submit()`.

internal/data/payments.go
```diff
@@ -663,6 +700,11 @@ func newPaymentQuery(baseQuery string, queryParams *QueryParams, sqlExec db.SQLE
 	if queryParams.Filters[FilterKeyStatus] != nil {
 		addArrayOrSingleCondition[PaymentStatus](qb, "p.status", queryParams.Filters[FilterKeyStatus])
 	}
+	if walletIDs, ok := queryParams.Filters[FilterKeySourceWalletIDs].([]string); ok {
+		// Membership-filtered visibility. Direct payments (no disbursement) carry no
+		// explicit source wallet until and inherit default-wallet visibility.
```
hypekostas: Commit 51a8a9a3 ("drop internal W1-W4 workstream labels") has left this comment incomplete.

internal/serve/httphandler/payments_handler.go
```diff
@@ -87,6 +87,33 @@ func (p PaymentsHandler) GetPayment(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
 	payment, err := p.Models.Payment.Get(ctx, paymentID, p.DBConnectionPool)
+	if err == nil {
+		// Membership-filtered visibility: 404 outside the caller's scope — existence
+		// is never disclosed. Direct payments inherit default-wallet visibility until .
```
hypekostas: Commit 51a8a9a3 ("drop internal W1-W4 workstream labels") has left this comment incomplete.

internal/data/receivers.go
```diff
@@ -277,6 +277,11 @@ func newReceiverQuery(baseQuery string, queryParams *QueryParams, sqlExec db.SQL
 	if queryParams.Filters[FilterKeyCreatedAtBefore] != nil {
 		qb.AddCondition("r.created_at <= ?", queryParams.Filters[FilterKeyCreatedAtBefore])
 	}
+	if walletIDs, ok := queryParams.Filters[FilterKeySourceWalletIDs].([]string); ok {
+		// Membership-filtered visibility (, accepted flag R1): a receiver is visible to
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

internal/serve/httphandler/export_handler.go
```diff
@@ -34,6 +36,16 @@ func (e ExportHandler) ExportDisbursements(rw http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	// Membership-filtered visibility (, flag R4): exports carry only the caller's wallets.
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

internal/serve/httphandler/export_handler.go
```diff
@@ -88,6 +100,16 @@ func (e ExportHandler) ExportPayments(rw http.ResponseWriter, r *http.Request) {
...
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

internal/serve/httphandler/export_handler.go
```diff
@@ -197,6 +219,16 @@ func (e ExportHandler) ExportReceivers(rw http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	// Membership-filtered visibility (, flags R1/R4).
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

internal/scheduler/jobs/event_delivery_job.go
```diff
@@ -0,0 +1,128 @@
...
+	eventDeliveryMaxAttempts = 20
+	eventDeliveryHTTPTimeout = 10 * time.Second
+	// eventDeliveryRetentionDays bounds outbox growth: DELIVERED events older than this are
+	// pruned each run. Undelivered (including poisoned) events are NEVER pruned — they stay
+	// visible for operations until handled.
+	eventDeliveryRetentionDays = 30
+)
+
+// eventDeliveryJob delivers undelivered outbox events to the tenant's configured webhook URL
+// (, accepted spec flags E1/E3): at-least-once, oldest first; consumers deduplicate on
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

db/migrations/sdp-migrations/2026-06-09.7-create-events-table.sql
```diff
@@ -0,0 +1,29 @@
+-- Transactional outbox for webhook/internal events (, accepted spec flag E1): event rows are
```
hypekostas: Consider removing references like R1/R4, E1/E3, etc.

helmchart/sdp/dashboards/dashboard.json
hypekostas, line 1: Now the 3 dashboard JSONs are duplicated (~5,270 lines) between `resources/grafana/` and `helmchart/sdp/dashboards/`. The reason is understood (Helm's `.Files` can't read outside the chart), but are they necessary in both places? Future changes would have to be made to both to ensure they stay in sync.

internal/serve/httphandler/distribution_wallets_handler.go
```diff
@@ -0,0 +1,423 @@
...
+type WalletBalanceResponse struct {
+	WalletID string            `json:"wallet_id,omitempty"`
+	Balances map[string]string `json:"balances"`
+}
+
+// GetDistributionWalletBalance returns one wallet's live on-chain balances. Reads follow the
+// membership taxonomy: 404 outside the caller's scope (existence undisclosed).
+func (h DistributionWalletsHandler) GetDistributionWalletBalance(rw http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	walletID := chi.URLParam(req, "id")
```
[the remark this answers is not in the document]
hypekostas: This can be taken care of at a later stage.

xBalbinus:
Thanks for the thorough review — addressed everything in [`e9f4cf0`](https://github.com/xBalbinus/stellar-disbursement-platform-backend/commit/e9f4cf026b2e51d22b17e91f90e5db83dface503) (single commit, pushed to `feat/multi-wallet`).

**Fixed:**
- Wallet read-visibility bug in payments — was falling back to the default wallet instead of the payment's actual `source_wallet_id`, leaking/hiding cross-wallet data
- `PostDirectPayment`'s balance/trustline check now validates the routed wallet, not the tenant's legacy account
- TSS signing (`buildAndSignTransaction` / `restoreArchivedEntries`) now resolves the wallet-specific distribution account — fixed both instances, including a second occurrence not explicitly called out
- API keys created by a since-deactivated user now return 401 instead of 500
- New non-owner users get a default-wallet membership on creation (previously silent empty-screens)
- `CreateWallet` no longer exposes an unfunded wallet as `ACTIVE` — added a `PENDING` status, promoted only after on-chain funding succeeds
- Closed the archive-trigger race (`FOR SHARE` on the guard triggers) — verified with a real two-transaction lock-timeout test, not just a code read
- Webhook outbox: added the missing write path, mandatory HTTPS, HMAC-SHA256 signing, and an SSRF guard on the delivery client — verified it actually blocks a loopback target
- Removed the demo-only Railway CI files; cleaned up dangling/stale comments

**Confirmed, not changed:**
- The replica row-reservation note on `event_delivery_job.go` — you flagged it low-priority; not applicable at current deployment scale
- `derive_payment_source_wallet`'s defensive-overwrite — you confirmed no live caller path reaches it; left the trigger as-is rather than modify an immutability guarantee without a real bug behind it
- OpenAPI spec update — you already called this for a later stage

**On the dashboard JSON duplication question:** verified a symlink (`helmchart/sdp/dashboards → ../../resources/grafana`) works cleanly end-to-end with `helm lint`/`template`/`package`. Happy to apply it if you'd rather not carry the duplication — otherwise a CI diff-check between the two directories is the lighter alternative.

Full test suite green across every package (50/50), with new regression coverage added for each fix.

hypekostas:
Another issue that will likely be caught by integration tests upon merging: **Circle tenants can no longer start disbursements**

Starting a disbursement now reads the paying account from the new `distribution_wallets` table(`disbursement_handler.go:659-673`) instead of resolving it the old way. That table only has a column for a Stellar address, and there's nowhere to store a Circle wallet ID, so Circle tenants break twice over. The previous code resolved via `DistributionAccountResolver.DistributionAccountFromContext` (`disbursement_handler.go:544-551` at `e5adbdff`), which fills in `CircleWalletID` from `circle_client_config`.

- They get a blank address. Circle tenants have no Stellar address by design: `provisionDistributionAccount` sets the type and `PENDING_USER_ACTIVATION` and returns without one (`manager.go:293-296`). `syncDefaultDistributionWallet` then only copies an address across for Stellar tenants (`manager.go:199-201`), so the row is created blank and `:664` rejects the disbursement with "the disbursement's source wallet has no funded distribution account yet".

- Circle setup doesn't fix it. When a tenant pastes in their Circle credentials, `circle_config_handler.go:116-119` updates the tenant record only. It never touches `distribution_wallets`, and this PR doesn't change that. So the tenant record goes ACTIVE while the wallet row stays blank and PENDING forever.

- Their wallet ID is lost. Even past that check, the account object built at `:669` sets only Address, Type and Status. The old path also filled in `CircleWalletID`, which is how Circle payments actually find the money. There's no column for it now, so it comes through empty.

Unit tests won't catch it either. `disbursement_handler_test.go` has no Circle cases at all, and the one test reaching this code asserts the blank-address rejection as correct (`:1943`) before filling in an address so the rest of the sub-tests can run (`:1948-1955`).

Simplest fix: fall back to the old resolution when the wallet isn't `STELLAR.DB_VAULT`. 
Better fix: give `distribution_wallets` a `circle_wallet_id` column and have the Circle config handler update the row when the tenant completes setup.

xBalbinus:
Superseded by #1178.

Same branch, same commits — re-opened from `feat/multi-wallet` in this repository rather than from a fork, so that CI can actually run.

Two things were blocked while this PR was fork-based:

- `db-migration` and `internal-sep-validation` could never pass. GitHub withholds Actions variables from fork PRs, so both containers exited at startup with `Invalid config: distribution-public-key is blank` / `sep10-signing-public-key is blank`. The variables were configured correctly all along — the same two jobs pass on in-repo PRs (#1162, #1139, #1119).
- Every workflow run needed manual maintainer approval, per commit.

Both are resolved on #1178.

@hypekostas — apologies for the move mid-review. Your threads here stay readable, and I've kept a written record of all of them; the fixes for every one are in the commits now on #1178. Happy to re-post any specific thread there if that's easier for you to track.

